PR #3, opened 2015-09-03
Add ability to make titles dynamic

tylermcginnis:
In our app, all of the navbar titles aren't static. We have a few places where we're putting the users name or the title of the specific post in the title. This pull request makes it possible to dynamically update the routes title by passing in the name of the title as the second argument to ```Animation.myRoute```. 

rfbis7azsv

The code for the above Gif looks like this,

```
    Actions.post({
      authedUser: this.props.authedUser,
      post: this.props.post
    }, this.getTitle(this.props.post.title));
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -104,9 +104,13 @@ class Router extends React.Component {
                     self.setState({initialRoute: child.props});
                 }
                 if (!(RouterActions[name])) {
-                    RouterActions[name] = function (data) {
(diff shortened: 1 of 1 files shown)


aksonov:
I believe it should be supported already. Have you tried to just pass title param to action?

Pavel.

tylermcginnis:
That works! This router is about 1000X better than the other one we were using. I really appreciate it!

Issue #4, opened 2015-09-04
Breaks when minified


jacob-israel-turner:
When I bundle/minify my app, this library breaks.  For some reason, the component names don't get added to the `Actions` object after minification.

`Object.keys(Actions)` prior to minification:

``` javascript
["push", "PUSH", "pop", "POP", "dismiss", "DISMISS", "preSplash", "splash", "login", "register", "etc."]
```

`Object.keys(Actions)` after minifcation:

``` javascript
[ 'push', 'PUSH', 'pop', 'POP', 'dismiss', 'DISMISS' ]
```

aksonov:
Thanks, it seems to be real issue - probably because Actions are generated in runtime, not compile time and minificator brakes it somehow. But it seems to be out of my control, any idea how to make it work with minification?

jacob-israel-turner:
I'll take a look at it this afternoon, I honestly have no idea how your library works yet. :P I'll let ya know what I find.

aksonov:
It works in straightforward way: it just adds new functions during processing of «Route» children to alt Actions object. Maybe I need to ask within alt github how to add custom actions in runtime correctly.

aksonov:
Please check latest version 0.1.0, this bug should be fixed.

jacob-israel-turner:
@aksonov Thanks.  We ended up going with an unminified bundle because of time restraints.  I'll take a look at 0.1.0 for our next release.

PR #5, opened 2015-09-04
Add string names to avoid obfuscation

mull:
Uglify tends to obfuscate class names. Adding a string to the
alt.createXXX methods tends to help.

Hopefully helps with https://github.com/aksonov/react-native-router-flux/issues/4

### DIFF
--- a/FetchActions.js
+++ b/FetchActions.js
@@ -1,6 +1,6 @@
 var alt = require('./alt');
 
-class Actions {
+class FetchActions {
     fetch(route, data){
         this.dispatch({route, data});
     }
@@ -23,4 +23,4 @@ class Actions {
 }
 
 
-module.exports = alt.createActions(Actions);
+module.exports = alt.createActions(FetchActions, "FetchActions");
--- a/actions.js
+++ b/actions.js
@@ -7,4 +7,4 @@ class Actions {
     }
 }
 
-module.exports = alt.createActions(Actions);
+module.exports = alt.createActions(Actions, "Actions");
--- a/store.js
+++ b/store.js
@@ -52,4 +52,4 @@ class RouterStore {
 }
 
 
-module.exports = alt.createStore(RouterStore);
+module.exports = alt.createStore(RouterStore, "RouterStore");


aksonov:
It doesn't work (with both unminification & minification)
ios simulator screen shot 4 2015 14 27 50

mull:
For some reason I thought the issue was only after minification. Gimme a couple of minutes/hours :)

aksonov:
Yes, 0.0.5 works well without minification, but 0.0.7 (with your PR) doesn't work at all. maybe I've broken something by myself with 0.0.6, let me check too.

mull:
FYI I'm getting complaints about this package being missing: https://www.npmjs.com/package/absolute-path

mull:
```
➜  Example git:(feature/minification-issues) ✗ npm start

> Example@0.0.1 start /Users/eml/Projects/react-native-router-flux/Example
> node_modules/react-native/packager/packager.sh


Could not find dependencies.
Ensure dependencies are installed - run 'npm install' from project root.
````

After adding the missing dependencies `absolute-path` and `chalk`

**Comment by aksonov** [comment]
Strange, are you using io.js?

**Comment by mull** [comment]
Ah that was indeed the case. Tried on `0.12.7` but got these errors:

```
React packager ready.

[2:40:46 PM] <START> fs crawl
TypeError: Cannot read property 'root' of null
    at /Users/eml/Projects/react-native-router-flux/Example/node_modules/react-native/packager/react-packager/src/DependencyResolver/crawlers/index.js:16:84
    at tryCallOne (/Users/eml/Projects/react-native-router-flux/Example/node_modules/react-native/node_modules/promise/lib/core.js:37:12)
    at /Users/eml/Projects/react-native-router-flux/Example/node_modules/react-native/node_modules/promise/lib/core.js:103:15
    at flush (/Users/eml/Projects/react-native-router-flux/Example/node_modules/react-native/node_modules/promise/node_modules/asap/raw.js:50:29)
    at process._tickCallback (node.js:355:11)
[2:41:02 PM] <START> request:/index.ios.bundle
[2:41:02 PM] <START> find dependencies
```

**Comment by mull** [comment]
Let me know if you introduced some trouble in 0.0.6. I need to get back to work but I can try more tonight or tomorrow morning. I was hoping my quick PR could've done it.

**Comment by aksonov** [comment]
You need to have io.js, check this:
http://facebook.github.io/react-native/docs/getting-started.html#content

No, problem only after your PR :) 
Read-only error happens because of second name parameter was added. Maybe after that i can't add new actions to class, so the question how to do it.

**Comment by aksonov** [comment]
Ok, i've rollback changes and released 0.0.8 to avoid errors for a while.

**Comment by mull** [comment]
I was on iojs 2.5 though. I'll follow the instructions later on. :)

That's really strange, it appears there's an API change in alt that I'm not aware of. Could you try going back to alt 0.17.1?

**Comment by aksonov** [comment]
I’m using 0.17.1 version of alt.

**Comment by mull** [comment]
Can you verify that is the one installed? Because running `npm install` would install `0.17.3`.

**Comment by aksonov** [comment]
You are right, 0.17.3 was installed, i've changed it to 0.17.1, but read-only error still exists.

**Comment by mull** [comment]
Thanks. I'll try and help you with this over the weekend. If we're lucky one of the other collaborators who aren't new and have better answers for you. :)

**Comment by aksonov** [comment]
Thanks, actually I agree with read-only error - way I'm adding custom actions are 'hacky' - i'm modifying internal properties of class now, but i didn't found API for that...

**Comment by mull** [comment]
Agreed, they are kind of hacky. A nicer way (to me) would be to dynamically create the action once you have your schema. But then that gives you issues when you import components that imports the action class before the schema is even defined.

**Comment by aksonov** [comment]
I think we really need API for alt to create actions runtime

Pavel.

**Comment by mull** [comment]
@aksonov you *can* create actions runtime. It's just that once you say `alt.createActions` the return value is immutable. That's not really a bad thing. Thinking about your problem during the evening I don't think mutable actions are the way to go here. After your schema definition you have a list of actions for your class, you could feed that into alt.createActions!

**Comment by aksonov** [comment]
Yes, I was thinking about it, but how to unite then my 'generated' actions with already existed actions (push/pop) - it will be not good to have two different classes for this? Just merge? Sorry, i'm  still learning Javascript, so could miss something obvious here.

**Comment by aksonov** [comment]
Oh, maybe i could just hide original actions (push/pop) inside component and just duplicate these methods in 'generated' actions ?

**Comment by mull** [comment]
Once you have your route definitions and <Schema> you could build the specification for the action & call alt.generateActions. You *could* though I don't know if this is a great solution pass that down through context to the react components. They could call `this.context.router.login()`.

**Comment by aksonov** [comment]
Usage of context will make usage more complex. It looks interesting idea - generate actions by its names (so it will be real actions), subscribe to all of them and handle properly within store. The problem is how to send these generated actions to user? Currently i'm just 'attaching' to actions singleton so user can easily import them as usual actions during _compile_ time and use it later, in runtime.

**Comment by mull** [comment]
Indeed it will.

Okay so for generating the actual action is what I would do. Get an array of action names such as `["login", "about", "contact", "faq", "tos", "feed"]` and use it to create your an object like:

```
{
  login: function(data) { this.push("login", data),
  about: function(data) { this.push("about", data),
}
```

That can in turn be sent to `alt.createActions`- see http://alt.js.org/docs/createActions/. You want to merge in the implementation of `push` and what have you into that object though.

Once you have your object you can pass that `alt.createActions`and use that return value in your components. Does that make sense?

**Comment by aksonov** [comment]
I think in the same way, but how i can use it as return value for my module.exports, if it is done during compile time? 

Btw, i've commented all actions modifications within index.js and still get that read-only error for 'named' actions.

**Comment by mull** [comment]
You can't use it as return value for your module.exports. You'd have to modify that value when you've generated your action.

```
let _action = {};

export function _setAction(value) { _action = value; }
export default _action;
```

Pretty sure any file that imports the file will have the updated value. Given that you should be doing the whole config & generation before any component is ever rendered and has a chance to call your action, it would work.

aksonov:
Hm, interesting, is it feature of ES6? I.e. if i call _setAction from my component, it will update _action for all other files imports it? 
Looks like current react-native doesn't support ES6 fully, but only some feature (like 'class' declaration)

ios simulator screen shot 4 2015 22 35 37

mull:
Yeah that would be an ES6 feature. Typically in ES6 projects you run the code through Babel to handle this kind of stuff. If you wish to pursue this path take a look at babel. This looks good: http://daliwa.li/es6-modules-in-node/

aksonov:
Unfortunately that doesn't work within React Native environment - i don't want to patch run script or something like that because this way module will be unusable. I will try to find other solution

aksonov:
@mull Okey, i've implemented some solution without modifying existing actions anymore but generate own ones, but result is the same - it works without minification, but doesn't work with.

As I mentioned above - when I'm applying your PR (adding names for actions), i'm getting readonly error always, even without any modification of actions.

aksonov:
I've fixed bug with minification #4, it happens in code unrelated to alt.

PR #6, opened 2015-09-10
make components transparent

markburns:
I have a router in front of a background so I wanted to be able to see through to the background.
Obviously any components can set their background to be white or whatever to not be transparent, but the opposite is not true if the default is white.

Let me know what you think.


aksonov:
Looks good!

PR #18, opened 2015-10-12
adding refs for navBar and footer

FaridSafi:
This way we can access to the navbar from an action using : this.props.navigator.refs.navBar

### DIFF
--- a/index.js
+++ b/index.js
@@ -157,13 +157,15 @@ class Router extends React.Component {
         if (navBar) {
             navBar = React.addons.cloneWithProps(navBar, {
                 navigator: navigator,
-                route: route
+                route: route,
+                ref: 'navBar'
             });
         }
         if (footer){
             footer = React.addons.cloneWithProps(footer, {
                 navigator: navigator,
-                route: route
+                route: route,
+                ref: 'footer'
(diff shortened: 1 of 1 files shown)


aksonov:
Still don't see how it could be accessed via name 'navBar'? Also why you need it?

FaridSafi:
Thank you for your answer.

By clicking on the "Next" button of the navbar, I want to retrieve the values of the switches of the page (see attached screenshot).
Since the navbar is created before the switches, I can not pass an onNext event containing the switches references when calling the action.

The only working way I found is to access this.props.navigator.refs.navBar inside the action.
Do you think I should use another way?


simulator screen shot 12 oct 2015 19 22 41

aksonov:
Your idea is not "Flux-like". All UI layer should do is just calling actions, UI should not have any logic.
You should define actions for each field and store which will listen them and change its state. "Next" will call some another action as well and that store will listen it and process it without problems - that store will have all values already, right?

FaridSafi:
Ok I understand now, thank you
I followed demo https://github.com/mrblueblue/react-native-alt-demo to store the values of the fields

PR #31, opened 2015-11-03
use alt's magic method

daemonchen:
use alt's magic method

### DIFF
--- a/store.js
+++ b/store.js
@@ -18,12 +18,13 @@ class RouterStore {
     constructor(){
         this.routes = [];
         this.currentRoute = null;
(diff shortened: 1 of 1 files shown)


aksonov:
Will it work if some actions are not handled, like 'switch' action inside store.js ?

daemonchen:
In alt, there is no switch statement, does it ?

aksonov:
Sorry, i meant ContainerStore which adds 'switch' action. Also user may add 'custom' actions, so store.js cannot handle all actions.

daemonchen:
Oh, I think it can handle all user's 'custom' action in 'onCustom' way.

PR #57, opened 2015-12-11
demoting Button to devDependencies

herby:
`Button` is not used in the main file, only in an example.
Removed the import in the main file, and changing the dependency to be dev-only.


aksonov:
Thanks, did it manually, because of conflict - already modified code during fixing other issues

PR #77, opened 2015-12-15
PR to pass styles down to tabs  

iZaL:
pass props to tabs tabStyle={{backgroundColor:'green'}}


aksonov:
iZal, thanks, but could you do this PR for react-native-tabs? I want to make this component more abstract from tabs representation...

iZaL:
I just read react-native-tabs api and it seems that that the option to pass styles down to tabs is already there, **iconStyle** prop does exactly the same thing what this PR does, so this PR is obsolete ..

Issue #56, opened 2015-12-11
Documentation example on how to integrate with a redux/flux architecture


jmacedoit:
"The component doesn't depend from any Flux implementation and allows to intercept all route actions by adding Actions.onPush/onReplace/onPop handlers from your store(s)"

I think an example on how to accomplish this would be very valuable. From inspecting the code, I have an idea on how I could do it, but I'm not sure if it's the proper way.

aksonov:
The problem that for now i don't see use cases to use onPush/onReplace/onPop, maybe you could provide some use cases when it is needed?

jmacedoit:
Perhaps a "log out" case where you need to manipulate some data and pop?

aksonov:
Again, i don't see how onPop could be used here. You just executes two actions, one - logical, 'logout' for your flux/redux which will manipulate all needed data and doing pop at the same time...

<Button onPress={()=>{Actions.pop(),this.dispatch(logout())} }>Logout</Button>

herby:
Yeah. just some hint, please. I just don't know how to combine that with a state.

aksonov:
I believe info passed via react context. Check it http://rackt.org/redux/docs/basics/UsageWithReact.html
I'm just using these two steps and everything works well (root Provider, then connect each screen).
Maybe you could do redux demo, it would be really great.

herby:
I have read though the that page already, thanks. I understood how to connect things there, but there, root (in App.js) was connected, and it was passed state to `select` function, where it selected relevant subsets and used then directly in the render JSX in the components.

But here, those components are not specified at all. There are only routes. I do not understand how the data gets passed to the 'screen' components themselves (or do I add it to the `Route` and it gets passed down to the component?)

aksonov:
For Route you define 'component' class, that class should be 'connected'. Data passed via inner context, as i understand.

herby:
Ok, I probably understand now. I use `Provider` to specify store for an app, and then I need to connect every single 'visual' component, adding specific select function for each of it (so every one select its own relevant subset), and if I have hierarachy of visual components, then I connect the roots of the trees in the visual forest (and passing the rest down via props as shown in other redux-react tutorials). Am I right @aksonov?

aksonov:
Right

herby:
Thanks. ;-)

jmacedoit:
[...]

PR #86, opened 2015-12-19
Improved documentation

emrosenf:
- Cleaned up language and formatting
- Added information about known limitations
- Added details about element parameters
- Changed example code to use ES2015 imports

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,36 +1,71 @@
-# react-native-router-flux
+# React Native Router [![Join the chat at https://gitter.im/aksonov/react-native-router-flux](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/aksonov/react-native-router-flux?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-[![Join the chat at https://gitter.im/aksonov/react-native-router-flux](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/aksonov/react-native-router-flux?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-React Native Router Actions based on exNavigator (https://github.com/exponentjs/ex-navigator)
+Router for React Native based on [exNavigator](https://github.com/exponentjs/ex-navigator)
 
-## Why?
-- Use Actions to replace/push/pop screens with easy syntax like Actions.login for navigation to login screen
-- Forget about passing navigator object to all React elements, use actions from anywhere in your UI code.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,16 +146,38 @@ var styles = StyleSheet.create({
...
+
+For example, instead of 
+```
+<Route name="register" component={Register} title="Register"/>
+```
+
+you might write:
+
+```
+<Route name="register" component={connect(selectFnForRegister)(Register)} title="Register"/>
```
MITDelian: Is this really a common practice? I feel like it's more common to call connect in the component file, not inside the routing file.

MITDelian:
Looks pretty good! I think we could likely improve the formatting of the `Feature` section but this is a great step in the right direction

README.md
```diff
@@ -114,16 +146,38 @@ var styles = StyleSheet.create({
...
-    * If some your Routes have common attributes, you may define Schema element and just use 'schema' attribute for 'route'
-3. In any app screen:
-    * var {Actions} = require('react-native-router-flux');
-    * Actions.ACTION_NAME(PARAMS) will call appropriate action and params will be passed to the route
+## Redux/Flux support
+This component is not opinionated and does not depend on any implementation of Flux or Redux.
+
+All route actions can be hooked by adding handlers for `Actions.onPush`, `Actions.onReplace`, `Actions.onPop` in your store(s).
+
+If a handler returns false route action is ignored. For Redux, you will need to 'connect' your component to your store.
```
MITDelian: If a handler returns false *the* route action is ignored?

README.md
emrosenf: I took that from #56. Will you write up an example of how you are doing it?

MITDelian: Yeah I was just reading through that discussion

MITDelian: The way I've been doing it, which I think is more common practice than using `connect` inline has been like this:

```
export default connect((state) => ({
  email: state.data.ui.login.email,
  password: state.data.ui.login.password,
  status: state.data.ui.login.status,
  error: state.data.ui.login.error
}))(Login);
```

MITDelian: and then when you import the `Login` component it is already `connect`ed

README.md
emrosenf: Yes, look at index.js#L133

MITDelian: sorry that was more of a grammatical comment. I think the word `the` should be inserted in that sentence is all

MITDelian: `If a handler returns the false route action is ignored.`

emrosenf:
Fixed the typo, thanks

PR #118, opened 2016-01-12
Allow for icon style and text style to be passed

MITDelian:


MITDelian:
I was messing around with customizing the back button and realized that `this._barButtonIconStyle` and `this._barButtonTextStyle` are always undefined

PR #134, opened 2016-01-19
Added popToTop() in the onReplace method

rgoldiez:
Added `this.refs.nav.popToTop()` to the onReplace method of `ExRouter.js` so that the components **unmount** when replacing the current route with another route.  I was having an issue where one of my child routes that requires an event listener wasn't having it's `componentWillUnmount` called when replacing the route.  This change is working for my use case.

I'm new to this package so there may be a better way to solve this.

### DIFF
--- a/ExRouter.js
+++ b/ExRouter.js
@@ -202,6 +202,7 @@ export default class ExRouter extends React.Component {
                 return false;
             }
         }
+        this.refs.nav.popToTop();
         this.refs.nav.replace(new ExRouteAdapter(route, props));
         return true;
     }


rgoldiez:
Is unmounting all scenes not the expected behavior on a replace?  It is in my cases because the navigation stack could have scenes in the stack that have listeners for remote API events to finish.  More specifically, a scene that has listeners may not be the last / top event in the navigator stack.

aksonov:
So why don't use reset action then?

rgoldiez:
I am willing to try that, but I didn't see anything in the docs about reset.  How do I call that (Actions.reset() throws a not found error)?

aksonov:
Have you tried Actions.reset(route_name) ?  PR about docs is always welcome

rgoldiez:
Not yet, but I will.  Can I pass in props with this specific call?  I need to call a route of type='replace' and pass props at the time of the call.

aksonov:
Please check it, if it doesn't pass with Actions.reset(route_name, params), you may fix it and post PR.

rgoldiez:
Here's what I get when trying to call Actions.reset() ... any idea?

screen shot 2016-01-21 at 7 22 49 am

aksonov:
Ops, probably you have to define Actions with type=‘reset’ and call Actions.ACTION_NAME(params)

rgoldiez:
That did it!  Thanks.  I will try to do a PR later today updating the README indicating type='reset'.

Thanks!

PR #154, opened 2016-01-27
Stop popping if there is nowhere to go

MITDelian:


MITDelian:
I had this issue where if I would very quickly tap the back button it would call Actions.pop() twice. This would sometimes result in there not being a parentRouter and it would throw an error so this was my fix

MITDelian:
Pop can get called twice but I don't think it should throw an error since it can be done just by fast tapping the back button

PR #163, opened 2016-01-28
Tap on tab to go to home

MITDelian:


MITDelian:
I notice that in most apps when you have a tab navigation system and you are deep inside one of the tabs, clicking on that tab when it is already actively selected returns you to the "home screen" of that tab. This is how I implemented it. Happy to put it into the example code as well

PR #173, opened 2016-02-04
Clears Actions.currentRouter when unmounting ExRouter if this is the …

### LINKED ISSUE #172: "Cannot read property 'push' of undefined" when mounting router for the second time
I have an app that uses navigation only in a small part of it and for the rest it relies on figuring what components to render by state, using redux.

Today I replaced NavigatorIOS for routers in the main place where I used navigation, and at first it seemed to work perfectly. But when I mounted for the second time the navigation flow, after coming from other part of the app that does not use routers, I got the following error:

simulator screen shot feb 4 2016 3 51 20 am

I created a simplified scenario where I can simulate the problem here: https://gist.github.com/danielweinmann/8ab67efdf51e773f4a59

To simulate the problem, just run the example and do the following:

- Click on "Next"
- Click on "Sign in"
- Click on "Sing out"
- Click on "Next" again.

I would gladly contribute with a fix for this problem, but I couldn't figure it out so far.

Anyone has any thought on it? @aksonov, can you enlighten me?

danielweinmann:
…current router's ExRouter instance. Fixes #172 

I hope it doesn't have undesired side effects ;) but I took care of only clearing it if it is the current router's ExRouter.

What do you think, @aksonov?

### DIFF
--- a/ExRouter.js
+++ b/ExRouter.js
@@ -187,6 +187,12 @@ export default class ExRouter extends React.Component {
         this.state = {};
     }
 
+    componentWillUnmount() {
+        if (this === Actions.currentRouter.delegate) {
+            Actions.currentRouter = null;
+        }
+    }
+
     onPush(route: Route, props:{ [key: string]: any}):boolean {
         if (this.props.onPush){
             const res = this.props.onPush(route, props);


MITDelian:
Yeah @danielweinmann @aksonov this works on solving the referencing issue but the refs on the plugin are still empty for some reason

PR #177, opened 2016-02-05
add props "hideTabBar" to "Route"

jungsoo108:
TabBar Router
--Navigation Router
----Route1
----Route2 - tab bar must hide 

add some code, for upside case.

use <Route hideTabBar={true} component={....}}/>

### DIFF
--- a/Actions.js
+++ b/Actions.js
@@ -48,7 +48,7 @@ class Actions {
     dispatchEvent(type :string, router: any) {
         currentStack = router._stack[router._stack.length-1];
         route = router.routes[currentStack];
-        _eventListener[type].forEach(function(val, key){
+        _eventListener[type] && _eventListener[type].forEach(function(val, key){
             val(route);
         });
     }
--- a/Actions.js
+++ b/Actions.js
@@ -2,6 +2,8 @@ import Route from './Route';
 import Router from './Router';
 import debug from './debug';
 
+var _eventListener = {};
+
 function isNumeric(n){
     return !isNaN(parseFloat(n)) && isFinite(n);
 }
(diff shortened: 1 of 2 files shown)


aksonov:
Could you give me use case? I don't want to make the component more complex. Currently component doesn't know anything about 'tab bar'. Why you can't just push from tab route defined from root router? This way tab bar should not be shown for that route.

aksonov:
Could you just fix it within TabBar:
The problem that route.props.hideTabBar is not passed for Actions call below. pass it and check how it works.
    onSelect(el){
        if (!Actions[el.props.name]){
            throw new Error("No action is defined for name="+el.props.name+" actions:"+JSON.stringify(Object.keys(Actions)));
        }
        if (el.props.selected && Actions[el.props.defaultRoute]) {
          Actions[el.props.defaultRoute]();
        } else {
          Actions[el.props.name]();
        }
        InteractionManager.runAfterInteractions(() =>
            this.setState({hideTabBar: el.props.hideTabBar}));
        return {selected: true};
    }

jungsoo108:
list view (tab bar not hidden), press row
-> push item detail view (tab bar not hidden), press show map button
-> push map view(tab bar hidden)

i try push map view in root router, and encount problem after pop.
after pop map view, Actions.currentRouter is root router not router of item detail view.
so Actions.pop() on item detail view not working.

i think hideTabBar props of Route is one of solution.

if that is not good, help me, give another solution.

aksonov:
@jungsoo108 I've already gave direction above - use something like Actions[el.props.name]({hideTabBar:el.props.hideTabBar})  within onSelect method, have you tried it?

aksonov:
Okey, I've fixed it by myself in way i've described above, please check (version 2.1.9)

jungsoo108:
today, i see your comment and code.
sorry for late feedback.

your commited code is good, and it is really helpful for my app
thank you

PR #185, opened 2016-02-08
Passes Router props to Route's on subsequent renders

### LINKED ISSUE #184: Router should pass its own props to Route's on subsequent renders?
I have the following scenario:

1. I have a Router with its routes inside a presentational component, which I want to be totally unaware of state.
2. In a few routes of this router, I need to receive new redux state from my container.
3. The container re-renders the Router with my new state as props.
4. But since Router is not a React Component, it does not re-render the routes and does not pass the new props to the component.

Here is a Gist with a simple example (in the example I try passing props both to Router and Route): https://gist.github.com/danielweinmann/09a2c980fe2fa405eb39

What I suggest is the following:

* Add a passProps prop to Router
* Change the following block on ExRouter to pass these props manually to the route: https://github.com/aksonov/react-native-router-flux/blob/master/ExRouter.js#L293

I'm gonna send a PR with this soon, probably tonight, but wanted to document the issue first ;)

danielweinmann:
This fixes #184 ;)

### DIFF
--- a/ExRouter.js
+++ b/ExRouter.js
@@ -290,10 +290,16 @@ export default class ExRouter extends React.Component {
         return (
             <View style={styles.transparent}>
                 {header}
-                <ExNavigator ref="nav" initialRouteStack={router.stack.map(route => new ExRouteAdapter(router.routes[route]))}
-                         style={styles.transparent}
-                         sceneStyle={{ paddingTop: 0, backgroundColor:'transparent' }}
-                         renderNavigationBar={props=><ExNavigationBar {...props} router={router}/>}
+                <ExNavigator
+                    ref="nav"
+                    initialRouteStack={router.stack.map(route => {
+                        const oldProps = router.routes[route].props
+                        router.routes[route].props = {...oldProps, ...this.props}
+                        return new ExRouteAdapter(router.routes[route])
+                    })}
+                    style={styles.transparent}
+                    sceneStyle={{ paddingTop: 0, backgroundColor:'transparent' }}
+                    renderNavigationBar={props=><ExNavigationBar {...props} router={router}/>}
                     {...this.props}
                 />
                 {footer}


aksonov:
Could you explain this change in more detail? It seems ExRouter already passes their props to Router and Router passes it to child Routes.

danielweinmann:
@aksonov, the problem is detailed in issue #184, and this gist reproduces it: https://gist.github.com/danielweinmann/09a2c980fe2fa405eb39

Actually, I believe its the opposite of what you said: Router (the default export in ReactRouter.js) passes its props to ExRouter on every render.

Also Router (now the one in Router.js) receives these props and passes them to every Route. But _only in its constructor_. Here it is, in _addRoute: https://github.com/aksonov/react-native-router-flux/blob/master/Router.js#L91

Router (the one in Router.js) is not a React Component, so it doesn't have ways to pass the props to every Route again. But this shouldn't be a problem, because this should be ExRouter's responsibility (or RouterIOS or any other adapter).

That's when the issue becomes clear: ExRouter receives the routes as attribute of the prop this.props.router, but this prop is not updated on subsequent renders, exactly because Router (the one in Router.js) is not a React Component.

So, when ExRouter renders every Route using this.props.router.stack, it will have outdated/static props on them. My change updates the props of the routes on ExRouter's this.props.router.stack on every render, which seems to be necessary both for flux or redux flows ;)

If you feel like putting my point to test, run my gist on both master and pass_router_props_on_subsequent_renders branches. At least for me, it works on pass_router_props_on_subsequent_renders and it does not work on master.

aksonov:
Okey, what about 'rebuilding' Router when ExRouter receives new props? As I understand it should be componentWillReceiveProps and recreate Router there? For me it is more logical comparing with rebuilding intialStack (but leaving Router the same) as done in your commit, right?

aksonov:
To be more exact, i'm proposing to improve ReactRouter and its componetWillReceiveProps

danielweinmann:
Looks good, @aksonov! I think we can remove ReactRouter's constructor and move createRouter to it's render or componentWillReceiveProps.

I just don't think I understand the whole package enough to predict side effects to that. What do you think?

danielweinmann:
Having it on render() avoids duplication, actually, because using componentWillReceiveProps means that we must keep the constructor.

aksonov:
We need to carefully test it to avoid any side effects. For example performance issues are possible - there is absolutely no sense to rebuilt router each render if no related stuff is changed.

aksonov:
Btw, I'm preparing redux demo (with some redux improvements) now and maybe it will be usable then to show use cases when it is needed (to update router props) on that demo.

danielweinmann:
Got it! Question: what do you think of depending on something like https://github.com/substack/node-deep-equal ?

I can include a custom deepEqual function to the code, to check for changes in props before rebuilding router, also, but if you're open to more dependency maybe it's a good idea.

aksonov:
I'm ok with node-deep-equal

danielweinmann:
It didn't work with the ReactRouter componentWillReceiveProps approach. Somehow ExRouteAdapter's renderScene is getting this as an outdated ExRouteAdapter.

What do you think of merging this PR and creating an issue to make this better in the future? I have to move on with my project, and since it works now and the other approach have this weird behaviour (probably something we'll figure out soon enough) I think it's better to refactor in the future.

I'm interested in contributing more with this package in my free time, tough. So I'll probably tackle this myself soon.

aksonov:
Are you sure that this PR will not impact performance ? I.e. initialStack will be rebuilt for any new prop (even not related to router)

danielweinmann:
I'm not sure. Probably it will affect performance if it receives unrelated props.

But I think passing props carefully the responsibility of the caller ;) anyway, this is the standard behaviour in React. If you pass new props to a component (say, Router) it _should_ rebuild.

If anyone complains about performance issues in this case, we'll probably have to tell them not to pass props around irresponsibly :) but of course, if they have good arguments we can always improve our code.

MITDelian:
This seems to cause wrapRouter to get passed down as well which is not useful. I'm not sure I fully understand the purpose of this PR

MITDelian:
In my project these lines actually causes the application to crash because it gets stuck in an infinite loop of Router -> Route -> Router -> Route rendering over and over

danielweinmann:
@MITDelian the purpose of this PR is to have a way to re-render the components with new props when a parent's state, and thus props passed to its children, have changed. If you try the following Gist you'll see that it is a basic react/redux pattern that is broken: https://gist.github.com/danielweinmann/09a2c980fe2fa405eb39 ;) maybe you'll know of other ways to accomplish my goal. If so, please let me know.

@aksonov and @MITDelian, I think a solution would be to filter wrapRouter and other props specific to the router, preventing them from being passed. What do you think of that?

danielweinmann:
@MITDelian, just wondering: why do you pass wrapRouter to Router. Shouldn't it be passed only to Route?

Issue #188, opened 2016-02-09
Confusing nomenclature with the term router


danielweinmann:
Currently, 3 things are called router:

1. Router class on Router.js
2. The default export in ReactRouter.js (the whole documentation uses Router when requiring the class).
3. The router prop passed to Router (the one in ReactRouter.js)

I believe we should change them to something like this:

1. Router.js should become BaseRouter.js or something that makes clear that it is not the component we're going to import and render.
2. ReactRouter.js should become Router and have Router as class name in it, for debugging purposes.
3. The router prop should become adapter or something that makes clear that it is the implementation of an "abstraction". Router in ReactRouter.js is kind of abstract: it depends on ExRouter or other "adaptors" to work. The term adaptor is what comes to mind, but maybe there is something else.

What do you think of that, @aksonov?

aksonov:
Maybe docs are unclear about it, here is explanation: Router is pure router (no react dependency). ReactRouter is adoptation for React. ExRouter is 'plugin' for ReactRouter.

danielweinmann:
Yeah, I understand that. I just think it's better to have different terms for each thing. I also think it means trouble when you have to document these kind of thing instead of making it evident on the code.

This line creates the main problem, in my view: https://github.com/aksonov/react-native-router-flux/blob/master/index.js#L4

We're calling the react adaptation internally ReactRouter but exporting it as Router. It becomes more troubling because we already have a Router class internally as well.

As for the prop "router" on ReactRouter, I think it should be called "plugin", then.

aksonov:
I agree with you - Router as component name was used for backward compatibility with prev. versions and for easy reading of route map. Maybe we really should rename current Router class to BaseRouter and ReactRouter to Router, PR will be welcome.

danielweinmann:
Great ;) I'll try to do this and update the README as I do this.

I don't know when, though. If anyone else wants to do it, please feel free. I'll check here before starting.

philipshurpik:
Hi!
I think possibly can be good idea to clean-up index folder.
Now it' really confusing.
As a proposal: move Schema, Route, TabBar, Actions, Router, Animations to maybe 'components' folder

About ReactRouter, EXRouter - we need example for it
And I also agree with idea of renaming

PR #190, opened 2016-02-10
Renames Router to BaseRouter and ReactRouter to Router, and Router 'router' prop to 'plugin'

danielweinmann:
This PR implements what we talked about in #188.


philipshurpik:
thanks!
but if it's all ok with Example now?  
As I understand it used old Router version?

possibly we need to examples - with BaseRouter and Router

philipshurpik:
also - like a proposal - it can be good idea to clean-up index folder.
Now it' really confusing.
As a proposal: move Schema, Route, TabBar, Actions, Router, Animations to maybe 'components' folder

And ExRouter for example into 'plugins' folder

danielweinmann:
@philipshurpik, the example uses Router because we were exporting ReactRouter as Router before. That's the good part of this refactoring: it doesn't change anything for the user of the package, only internally.

I like the ideia of having ExRouter and RouterIOS in a 'plugins' folder! :D the other classes I don't know where is a good place for them yet. @aksonov?

aksonov:
@danielweinmann Why git shows BaseRouter as new file? I thought it show rename from Router to BaseRouter? Did you use git mv command?

aksonov:
Also agree about moving to components all React components

danielweinmann:
Sorry, @aksonov, I used git commit -am :S my bad.

aksonov:
@danielweinmann Could you do it in correct way?

danielweinmann:
Created a new PR here: https://github.com/aksonov/react-native-router-flux/pull/196

PR #200, opened 2016-02-11
Fixes passing route props again

### LINKED ISSUE #184: Router should pass its own props to Route's on subsequent renders?

danielweinmann:
Fixes #184 once again

### DIFF
--- a/ExRouter.js
+++ b/ExRouter.js
@@ -294,7 +294,7 @@ export default class ExRouter extends React.Component {
                 {header}
                 <ExNavigator ref="nav" initialRouteStack={router.stack.map(route => {
                         const oldProps = router.routes[route].props
-                        router.routes[route].props = {...oldProps, ...this.props}
+                        router.routes[route].props = {...oldProps, ...parentProps(this.props)}
                         return new ExRouteAdapter(router.routes[route])
                     })}
                     style={styles.transparent}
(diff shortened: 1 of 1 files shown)


danielweinmann:
@aksonov, the last commit kinda proves the point I was making in my last large comment on #184.

In these methods, this.props is outdated and props is updated. I don't know why yet, but I'm pretty sure it has to do with the way BaseRouter stores it's routes and props. These values stay static there, since there is no react lifecycle there.

Don't get me wrong: I believe its great that BaseRouter is not a React Component. We'll just have to think more carefully about it's architecture and investigate these problems more thoroughly ;)

PR #201, opened 2016-02-11
Update info on Redux/Flux support

mmazzarolo:


aksonov:
Ops, "If a handler returns false, the route action is ignored. For Redux, you will need to 'connect' your component to your store. If 'dispatch' prop is passed to the router, it will be called with current route as route, name as route name and all route props, check Example for more details." - no clear what handler here...

Issue #229, opened 2016-02-20
Get route title when using renderTitle


axelander:
How do I get the title of a route when using the `renderTitle` property? I tried using below code but it doesn't work. 

```javascript
renderTitle={props => (<CustomTitle { ...props } />)}
```
When using the default title functionality it works as expected, I can pass the title property in the action to the next route. 
```javascript
Actions.login({ title: 'Hello user' })
```

aksonov:
route is passed as first argument, so you need to use route.title

axelander:
Hm, still doesn't work.
The part of my router that I'm having trouble with looks like this:
```javascript
<Route name="tabbar">
  <Router footer={TabBar} showNavigationBar={false}>
    <Route name="tab_menu" schema="tab" title="Sections" style={styles.route}>
      <Router>
        <Route name="menu" schema="firstView" component={MenuView} />
        <Route name="page" schema="secondView" component={PageView} />
        <Route name="article" schema="secondView" component={ArticleView} />
      </Router>
    </Route>
  </Router>
</Route>
```
So, when switching to the tab `tab_menu` the menu route is default.
When dispatching an action from MenuView to go to the page route like this:
`Actions.page({ title: 'Sports' })`
I get "Sections" as the route.props.title in my custom title.
If I remove the renderTitle prop from my scheme (also tried adding directly to the routes with the same result) I get "Sport" as the title. So it works when using the default title behaviour.

aksonov:
Could you debug? It maybe bug.

axelander:
Seems like route is not passed to renderTitle callback. I created a PR for ex-navigator to fix this.
https://github.com/exponentjs/ex-navigator/pull/97

aksonov:
Hm, why PR to ex-navigator? ex-navigator is not connected with routing from this component.
As far as i see renderTitle is binded to route within ExRouter:
```
        if (this.route.props.renderTitle){
            this.renderTitle = this.route.props.renderTitle.bind(this.route);
        }
```

axelander:
When passing in the route like in my PR you get an `ExRouteAdapter` as the first argument in `renderTitle`. When removing it, you get a `ExNavigator`. So I guess the `.bind` part is not working as expected? Do you have any ideas how this could be fixed within RNRF?

axelander:
So, I found the problem with the `.bind` and created a [PR](https://github.com/aksonov/react-native-router-flux/pull/245).

aksonov:
[...]

PR #245, opened 2016-02-23
Bind ExRouteAdapter instance to render callbacks

axelander:
https://github.com/aksonov/react-native-router-flux/issues/229

### DIFF
--- a/ExRouter.js
+++ b/ExRouter.js
@@ -30,6 +30,7 @@ export class ExRouteAdapter {
         }
         this.route = route;
         this.name = route.name;
+        if (props.title) this.route.title = props.title;
         this.title = props.title || route.title;
         if (!this.name){
             throw new Error("name is not defined for route");
@@ -40,7 +41,7 @@ export class ExRouteAdapter {
             this.renderRightButton = this.route.props.renderRightButton.bind(this.route);
         }
         if (this.route.props.renderTitle){
-            this.renderTitle = this.route.props.renderTitle.bind(this.route);
+            this.renderTitle = this.route.props.renderTitle.bind(this, this.route);
         }
         if (this.route.props.renderLeftButton){
             this.renderLeftButton = this.route.props.renderLeftButton.bind(this.route);


aksonov:
Sorry, but it could brake existing apps. Why binding to this.route is not enough for rendering?

axelander:
I'm quite positive this is a bug but I'm not sure where to fix it.
I found that in the constructor of `ExRouteAdapter`, [on this line](https://github.com/aksonov/react-native-router-flux/blob/master/ExRouter.js#L33), title is set correct for the `ExRouteAdapter` instance but **not** the `Route`. Updating my PR to show you.
I really have no idea how this should be solved, you think you could help me?

ExRouter.js
```diff
@@ -40,7 +41,7 @@ export class ExRouteAdapter {
             this.renderRightButton = this.route.props.renderRightButton.bind(this.route);
         }
         if (this.route.props.renderTitle){
-            this.renderTitle = this.route.props.renderTitle.bind(this.route);
+            this.renderTitle = this.route.props.renderTitle.bind(this, this.route);
```
axelander: If route should be passed as the first argument to renderTitle then this update should be correct.

ExRouter.js
```diff
@@ -30,6 +30,7 @@ export class ExRouteAdapter {
         }
         this.route = route;
         this.name = route.name;
+        if (props.title) this.route.title = props.title;
```
axelander: Feels like I'm correcting something that should be handled earlier on in some other module, but this solves the issue with getting the parent route title.

aksonov: Below we have 
```
this.title = props.title || route.title;
```

aksonov:
Looks like the problem that Route doesn't update their props during actions. Then probably we have to fix BaseRouter. It should modify stack route with latest props, could you check there?

For example:
```
    _push(name:string, props:{ [key: string]: any} ){
        this._stack.push(name);
    }
```

Probably it should be added :
```
this.routes[name].props = Object.assign({}, this.routes[name].props, props);
```
What do you think? I don't have much time now to add such calls for all action types, check if everything is OK with demo, maybe someone could check.

aksonov:
@danielweinmann Could you look it? I still don't like your fix when you rebuild whole route stack for new props coming from Router, maybe it would be more correct to fix BaseRouter and set new props after each action there?

ExRouter.js
axelander: yes but this.title is on the `ExRouteAdapter` but the renderTitle callback is bound to `this.route`.

danielweinmann:
@aksonov I'll try to take a look later this week!

About your suggestion, I don't think after new action is a good enough place to set new props. It's important that every react component re-renders when receiving new, different, props.

No matter what they know about their own state, thinks can change on the outside and shouldn't have to dispatch an action to change.

I'm still guessing the main reason you guys don't share my line of thinking is because you all must be using a big root router that manages a whole app. Is that right? In my case, I have a lot of state management without any route and a few small routers deep inside my rendering stack. This makes sense for me, since I only want to have a router when I need navigation functionality, and in a big part of my app I don't need navigation. It would mean a lot on unnecessary code to have a big root router.

So I need a way to do just that: rebuild the whole stack for new props coming from Router.

PS: I still don't like my solution in terms of elegance of the code. But it solves this scenario, and any other solution presented so far don't.

PS2: I think when @MITDelian finishes making BaseRouter immutable we will see more clearly how to make my solution better.

aksonov:
Closing it for now because of 3.0 release is published.

PR #284, opened 2016-03-05
Use route props first when determining route title

jopecko:
When invoking Actions.refresh({title: 'Changed'}), the call to
super.render() in ExNavigationBar calls back into getTitle() on
ExRouteAdapter to determine the title to display in the navigation bar
This patch first checks the provided props before falling back to the
previous behavior.

### DIFF
--- a/ExRouter.js
+++ b/ExRouter.js
@@ -82,8 +82,9 @@ export class ExRouteAdapter {
     }
 
     getTitle() {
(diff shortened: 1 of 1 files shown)


farwayer:
You can swap `props.title` and `route.title` at https://github.com/aksonov/react-native-router-flux/blob/master/ExRouter.js#L33 to change title priority. It's more simple and correct way.

PR #358, opened 2016-03-21
Support deeply nested scenes setting TabBar properties

thessem:
If a scene is nested too deeply in a component with a tab bar it loses the ability to control the visibility of the tab bar.

This change makes the TabBar search down the list to the bottom of the scene until it hits the bottom.

Please advise if this needs to check the value of hideTabBar at every level, or if it only needs to search downwards until it finds a valid value or something

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -25,6 +25,9 @@ export default class extends Component {
     render(){
         const state = this.props.navigationState;
         const selected = state.children[state.index];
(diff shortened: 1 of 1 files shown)


aksonov:
@thessem Okey, where is PR?

thessem:
I decided to wait until I could put together a few other changes in the same style as this one, if you think this PR is acceptable by itself I'll happily reopen it though

PR #389, opened 2016-03-24
Added action type: 'pushToCurrent'

joenoon:
first finds the current scene's parent and sets it as the parent of the new scene.  Scenes defined at the root level can be pushed into tab scenes.
- pushed scenes are added to the root state and removed on pop
- break early in findElement
- change getCurrent to return entire scene

*This solves my current use case, but please check your use cases to make sure I haven't altered other expectations of the library.*

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -39,6 +39,8 @@ function getCurrent(state){
     return getCurrent(state.children[state.index]);
 }
 
+var _uniqPushed = 0;
+
 function update(state,action){
     if (!state.scenes[action.key]) {
         console.log("No scene for key="+action.key);
@@ -67,9 +69,11 @@ function update(state,action){
                 assert(el, "Cannot find element for parent="+el.parent+" within current state");
             }
             if (el.children.length > 1) {
-                el.children.pop();
+                let popped = el.children.pop();
                 el.index = el.children.length - 1;
-                delete newState.scenes[newState.scenes.current];
+                if (popped.ephemeral) {
+                    delete newState.scenes[popped.key];
+                }
                 newState.scenes.current = getCurrent(newState).key;
                 return newState;
             } else {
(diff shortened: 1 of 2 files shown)


src/Reducer.js
```diff
@@ -72,7 +69,8 @@ function update(state,action){
             if (el.children.length > 1) {
                 el.children.pop();
                 el.index = el.children.length - 1;
-                newState.scenes.current = getCurrent(newState);
+                delete newState.scenes[newState.scenes.current];
```
aksonov: Deleting looks not good - scenes should store all possible scenes, like library.

src/Reducer.js
```diff
@@ -86,10 +84,18 @@ function update(state,action){
...
+            el = findElement(newState, parent);
+            assert(el, "Cannot find element for parent="+parent+" within current state:"+JSON.stringify(newState));
+            // fall through to PUSH_ACTION
+
         case PUSH_ACTION:
             el.children.push(getInitialState(newProps, newState.scenes, el.children.length, action));
             el.index = el.children.length - 1;
-            newState.scenes.current = getCurrent(newState);
+            newState.scenes.current = getCurrent(newState).key;
+            newState.scenes[newState.scenes.current] = newProps;
```
aksonov: Modification of scenes could bring some unpleasant issues - it is possible that we could have several scenes in stack with the same sceneKey (such issue was created already in the past), for example on Details of user1 and then Details of user2. And you are proposing to set 'scene' with user1 or user2?

aksonov:
Please check my notes. Is it possible not to modify state.scenes?

joenoon:
Thanks for the notes, I'll dig into this!  The scenes I'm adding and deleting are what I was considering 'runtime scenes'.  The original defined scenes shouldn't be affected.  What I mean by this is this is what happens when pushing then popping some scenes:

'create scene', '1_List'
'create scene', '2_Profile'
'create scene', '3_List'

'delete scene', '3_List'
'delete scene', '2_Profile'
'delete scene', '1_List'

Those are internal/runtime scenes that were not recorded previously at the root level - I added them in this PR.

Let me think about this today and see if I missed something.

joenoon:
@aksonov I see the problem with my PR and the scene modifications... let me see if I can find another way.

joenoon:
Updated and open for comments!  I think this works and ensures the only things deleted from scenes are what were added at runtime, and also makes sure clones added are unique.

Also, I changed https://github.com/aksonov/react-native-router-flux/pull/389/files#diff-98b3594cf65eb9acfce1160c51678d6eR86 - I think this probably breaks the original case but fixes mine, so I need to look at this more.  But first I want to make sure the other stuff is making sense and on the right track.

joenoon:
This needs more work and isn't ready for merging yet.  I've discovered a few more problems with my PR related to how sceneKey and key were intended to be used. I need to get a better understanding of how everything fits together.  I'll keep working toward a PR over the next few days.

aksonov:
Closing it. Feel free to submit new PR for latest version. I hope it will be just setting parent to current one for pushToCurrent actions.

PR #400, opened 2016-03-26
clone option on Scene to dynamically push to current parent

joenoon:
Really excited for this one!  I hope the use-case is obvious from the example, but please let me know if I can clarify anything or if anything looks questionable.  The changes are isolated so should not affect existing behavior.

Also, I set the package.json react-native-router-flux to file:../ to make the example development a little easier going forward, but I can change that back if needed.

### DIFF
--- a/Example/Example.js
+++ b/Example/Example.js
@@ -7,6 +7,7 @@ import {Scene, Reducer, Router, Switch, TabBar, Modal, Schema, Actions} from 're
 import Error from './components/Error'
 import Home from './components/Home'
 import TabView from './components/TabView'
+import EchoView from './components/EchoView'
 
 class TabIcon extends React.Component {
     render(){
@@ -47,6 +48,7 @@ export default class Example extends React.Component {
         return <Router createReducer={reducerCreate} sceneStyle={{backgroundColor:'#F7F7F7'}}>
             <Scene key="modal" component={Modal} >
                 <Scene key="root" hideNavBar={true}>
+                    <Scene key="echo" clone component={EchoView} />
                     <Scene key="register" component={Register} title="Register"/>
                     <Scene key="register2" component={Register} title="Register2" duration={1}/>
                     <Scene key="home" component={Home} title="Replace" type="replace"/>
--- a/Example/components/EchoView.js
+++ b/Example/components/EchoView.js
@@ -1,0 +1,36 @@
+'use strict';
+
+var React = require('react-native');
+var {View, Text, StyleSheet} = React;
+var Button = require('react-native-button');
+var Actions = require('react-native-router-flux').Actions;
+
+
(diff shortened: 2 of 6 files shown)


aksonov:
Could we avoid modification of scenes? It will cause re-rending of whole nav tree. I don't understand why that 'cloned' Scene cannot have the same sceneKey as original scene. Because of parent? I believe for pushToCurrent you could just set parent before 'update'. Maybe i'm missing something please tell.

I also don't like addition of new attributes to Scene, could you just define new action type?

joenoon:
It could be that I'm missing something... lets figure it out.

Is it possible to get the behavior I added in the example without this `clone` feature?  My use-case is similar to many apps that have lots of interconnected content: Imagine a facebook newsfeed.  Each post has a creator, maybe target object (Place, Location) users who like it, comments.  Comments have a creator.  Location and Places might have their own list of users who like them, etc.  Lots of object types and very interconnected.  At any point I click on a user's name, a new page is "pushed" onto the navigation stack.  Then I look through that users followers and click on another user, which is "pushed" onto the navigation stack.  Then I look through that user's profile and list of places they like, and click a place.  The place page is then "pushed" onto the navigation stack.

That is what this PR is for: to be able to push certain scenes anywhere, anytime, multiple times, with the same or new attributes. (User 1 > Place 2 > Photo 3 > User 1 > User 5)

My understanding is without this feature, I would need to know the entire scene map at app launch.  Is that right?

aksonov:
So what the problem to create pushToCurrent action type, search 'current' parent and push it there?

aksonov:
Okey, i will merge and then I or you will try to improve.

joenoon:
@aksonov The problem I'm running into with 6d2e5d114118e594168f65c83732e8a470c0529f is, in my app there can be multiple pushed ProfileScreenComponents with different users as people click though the app.  But because they all share the same scene now, they step on each other during a refresh.

```
class ProfileScreenComponent extends React.Component {
  componentDidMount() {
    Actions.refresh({
      key: this.props.navigationState.sceneKey,
      rightTitle: this.props.user.name,
      onRight: () => (null)
    });
  }
  // ...
}
```

In this case, all ProfileScreen navbars rightTitle become the last user's name that was mounted, so if you go: Joe > Jane > Bob > Sarah > Jenn > Bill, and the start popping, "Bill" will be the rightTitle for every screen.

I know you didn't like the idea of adding scenes at runtime, but does anything else come to mind that would make this pattern work?  In my mind each unique profile screen should have a unique scene, but I'm not sure how to achieve that without adding scenes at runtime.

aksonov:
You need to fix refresh then to find current screen (with unique key) and set properties there, so you will be able to set rightTitle and onRight only for that screen....

PR #406, opened 2016-03-26
Added DrawerExample

Froelund:


aksonov:
It looks great, but could you add it to existing Example (tab bar screen) to switch tab1..tab5 ? I don't want to have two example projects because i will have to support both for any fixes/changes.

Froelund:
@aksonov I just added support to toggle the drawer in the default Navbar. I will merge the DrawerExample into the existing Example project.

aksonov:
Great!

aksonov:
@Froelund Sorry, it doesn't work properly for me - i see that button at start screen and all titles within tab bar are hidden (issue #421)

aksonov:
Improved Example with Drawer

PR #407, opened 2016-03-26
Allow upgrading to React Native 0.23

bosung90:


giacomorebonato:
[There is a bug](https://github.com/facebook/react-native/issues/6638) in react-native 0.23.0-rc1 that does not allow to require images and throws with react-native-router-flux because of the back arrow images

aksonov:
There is already request #423 for new version.

PR #411, opened 2016-03-27
Feature/override images

RWOverdijk:
Implemented a way to override the default images. I've tested this in my app and it works.

The reason for this PR, is because it's quite a lot of work atm to just override the image (you have to render the entire thing). This way, you can just do: 

```jsx
<Scene key="root" backButtonImage={require('image!back_chevron')}>
```

### DIFF
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -89,7 +89,7 @@ export default class NavBar extends React.Component {
               var drawer = this.context.drawer;
               drawer.toggle();
             }}>
-                <Image source={backButtonImage)} style={[styles.backButtonImage, this.props.navigationState.barButtonIconStyle]}/>
+                <Image source={backButtonImage} style={[styles.backButtonImage, this.props.navigationState.barButtonIconStyle]}/>
             </TouchableOpacity>
           );
         }else{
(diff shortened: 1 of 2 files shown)


src/NavBar.js
```diff
@@ -77,25 +77,33 @@ export default class NavBar extends React.Component {
...
+        if(!!this.context.drawer && typeof this.context.drawer.toggle === 'function'){
+
+          backButtonImage = this.props.navigationState.drawerImage || require('./menu_burger.png');
+
+          return (
+            <TouchableOpacity style={[styles.backButton, this.props.navigationState.leftButtonStyle]} onPress={() => {
+              var drawer = this.context.drawer;
+              drawer.toggle();
+            }}>
+                <Image source={backButtonImage} style={[styles.backButtonImage, this.props.navigationState.barButtonIconStyle]}/>
```
Froelund: Consider renaming backButtonImage to buttonImage, as it is used for both drawerImage and backButtonImage. For readability.

RWOverdijk: @Froelund Fair enough. See: https://github.com/aksonov/react-native-router-flux/pull/420

Issue #417, opened 2016-03-28
NavBar Title


ajoshdee:
Hi, after update to 3.2.3, i'm having trouble displaying the navbar title. My initial scene is a tab view. But the title is not showing. Inside the view is a button that will push another scene, In this scene I can see the title. I already tried hiding the root scene navbar. It will make the tab view's nav title visible but will make the pushed scene's navbar invisible. Any solutions?

philipshurpik:
I also faced with this problem and created pull request to fix it

philipshurpik:
as I understand now it's fixed, please check

aksonov:
Closing it for now, please report if it is not fixed in latest 3.2.4

PR #418, opened 2016-03-28
fix to show correctly show navbar in tabs component

philipshurpik:
Problem that when we have tabbar with tabs.
And tabs container has `hideNavBar={true}`
And root component has not `hideNavBar={true}` 
than tabs navbar is rendered (and it is empty)

Correct behaviour - only tabs headers are rendered 
also see issue # #417


ajoshdee:
Hi, how can I update to this ver. using the npm. I tried reinstalling it but the changes were not made.

PR #423, opened 2016-03-29
change DefaultRenderer to correspond new API

philipshurpik:
New version - to work with 0.23
But... 
Problem with icons still exists, so possibly couldn't be merged in master now

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-router-flux",
-  "version": "3.2.3",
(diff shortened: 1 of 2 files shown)


aksonov:
Yes, i've got error about AssetRegistry so i've reverted it back...

philipshurpik:
yes... i tested all other functionality - with this improvements it works on 0.23 (if just comment requiring images)
so when this error will be fixed - you can use it for next release

Issue #408, opened 2016-03-27
Upgrading to V3.1.3 has broken replace type


JamesWatling:
Since upgrading to the latest `v3.1.3` version the replace functionality is not working correctly.

I have updated my previous code to account for the new updates in V3, but the replace type acts as a normal push state. Have I simply overlooked something, or is this a new bug.

```
  <Scene
       key={Routes.LESSONS_INDEX}
       initial={this._loggedIn()}
       component={LessonsIndexContainer}
       type="replace"
   />
```

On another note, love what you're doing here, but due to the speed of releases do you think it would be wise to have a more formal changelog that details all the changes in each release.

Also if you require any assistance for updating the documentation let me know :smile:

marcosbeirigo:
Don't uses this terrible shit

aksonov:
I can do change logs, but it will make all things much much slower - so far many people want to use but nobody want to help.

Always check latest npm versions from NPM, not from github (because i can't create each version there)

PR #426, opened 2016-03-29
Properly replace entire history stack - Fix for Issue #408

nitsujri:
Current `type='replace'` allows going "back" through history still.

This replaces entire history stack.

https://github.com/aksonov/react-native-router-flux/issues/408


aksonov:
What you did is not 'replace' but 'reset' action. You may add it and submit PR (also add it to README)

PR #429, opened 2016-03-29
make refresh by key modify the child data and not scene data.

joenoon:
refresh now works with key and not sceneKey, making the previous workaround of preserving key and parent from state unnecessary.

@aksonov let me know what you think - happy to refactor or take another shot at this if anything looks off.


aksonov:
Looks ok, but could you unite getElement and getChildForKey? Like i did within inject:
```
const condition = action.type == REFRESH_ACTION ? state.key === props.key || state.sceneKey === action.key : state.sceneKey == props.parent;
 ```

joenoon:
@aksonov is this what you had in mind?

PR #435, opened 2016-03-30
hideNavBar hide scene only for current scene

philipshurpik:
As written in docs, hideNavBar "hides navigation bar for this scene", but previously it can hided it for children too.

As I understand - it's more correct and behaviour.
For example if we want to show beautiful flowFromBottom animation (when title animates with page) - that should work:
```
<Scene key="newPage" direction="vertical" hideNavBar={true}>
   <Scene key="newPageView"
	   title="New "
	   component={Page}
	   leftTitle="Cancel"
	   onLeft={props => Actions.pop()}
	   rightTitle="Done"
	   onRight={props => makeDone(props.id)}
    />			
</Scene>
```

### DIFF
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -63,7 +63,7 @@ export default class NavBar extends React.Component {
         if (selected.component && selected.component.renderNavigationBar){
             return selected.component.renderNavigationBar({...this.props,...selected});
         }
-        if (state.hideNavBar || child.hideNavBar || selected.hideNavBar){
+        if (child.hideNavBar || selected.hideNavBar){
             return null;
         }
         return (


aksonov:
I'm worrying it could broke existing apps with v3, looks like breaking change?

philipshurpik:
Yes, worst case scenario - navbar would be shown in places, when it was previously hidden by parent.

I think that this change will make usage of router more configurable.

Possibly it can be part of router 3.3 version with upgrade guide.
And can be introduced the same time, as RN 0.23 released.

P.S. Also we can make router to understand hideNavBar={false} to override parent behaviour, but it's not obvious solution

philipshurpik:
@aksonov Hi! :)
What do you think about this change now, but only for 3.24 branch?

aksonov:
I'm ok with it, will you submit PR for 3.24 branch?

philipshurpik:
yes, will make it in few minutes

PR #450, opened 2016-04-01
Added popToTop functionality

AlexDM0:
Hi,

In my app there are a few tabs, and the state is persisted if you toggle between them (thanks to PR #390). If you are already inside the tab however, I would like the state to be reset to the base of the tab for easy navigation. 

I've added an action called popToTop (name based on https://facebook.github.io/react-native/docs/navigator.html#content). 

I can now do this in my reducer:
```js
const reducerCreate = params=> {
  return (state, action)=>{
    if (state) {
      let currentTabIndex = state.children[0].index;
      if (state.children[0].children[currentTabIndex].name === action.key && action.type === 'jump') {
        return defaultReducer(state, {...action, ...{type:'popToTop'}});
      }
    }
  }
};
```

I figured I'd do a pull to see if you'd like to add it to the repo. I couldnt find where in the docs it would be added but if you can point me to it I can add it there too.

Regards

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -7,7 +7,7 @@
  *
  */
 
-import {PUSH_ACTION, POP_ACTION2, FOCUS_ACTION, JUMP_ACTION, INIT_ACTION, REPLACE_ACTION, RESET_ACTION, POP_ACTION, REFRESH_ACTION} from "./Actions";
+import {PUSH_ACTION, POP_ACTION2, POP_TO_TOP, FOCUS_ACTION, JUMP_ACTION, INIT_ACTION, REPLACE_ACTION, RESET_ACTION, POP_ACTION, REFRESH_ACTION} from "./Actions";
 import assert from "assert";
 import Immutable from "immutable";
 import {getInitialState} from "./State";
@@ -48,6 +48,11 @@ function inject(state, action, props, scenes) {
             case POP_ACTION:
                 assert(!state.tabs, "pop() operation cannot be run on tab bar (tabs=true)")
                 return {...state, index:state.index-1, children:state.children.slice(0, -1) };
+            case POP_TO_TOP:
+                assert(state.tabs, "Parent="+state.key+" is not tab bar, popToTop action is not valid");
+                let child = state.children[state.index];
+                let newChild = {...child, ...{children: [child.children[0]], index:0}};
+                return {...state, ...{children:[...state.children.slice(0,state.index), newChild, ...state.children.slice(state.index+1)]}};
             case REFRESH_ACTION:
                 return {...state, ...props};
             case PUSH_ACTION:
@@ -152,6 +157,7 @@ function reducer({initialState, scenes}){
         switch (action.type) {
             case POP_ACTION2:
             case POP_ACTION:
+            case POP_TO_TOP:
             case REFRESH_ACTION:
             case PUSH_ACTION:
             case JUMP_ACTION:
--- a/src/Actions.js
+++ b/src/Actions.js
@@ -13,6 +13,7 @@ export const PUSH_ACTION = "push";
 export const REPLACE_ACTION = "replace";
 export const POP_ACTION2 = "back";
 export const POP_ACTION = "BackAction";
+export const POP_TO_TOP = 'popToTop';
 export const REFRESH_ACTION = "refresh";
 export const RESET_ACTION = "reset";
 export const INIT_ACTION = "init";


src/Reducer.js
```diff
@@ -48,6 +48,11 @@ function inject(state, action, props, scenes) {
             case POP_ACTION:
                 assert(!state.tabs, "pop() operation cannot be run on tab bar (tabs=true)")
                 return {...state, index:state.index-1, children:state.children.slice(0, -1) };
+            case POP_TO_TOP:
+                assert(state.tabs, "Parent="+state.key+" is not tab bar, popToTop action is not valid");
+                let child = state.children[state.index];
+                let newChild = {...child, ...{children: [child.children[0]], index:0}};
+                return {...state, ...{children:[...state.children.slice(0,state.index), newChild, ...state.children.slice(state.index+1)]}};
```
aksonov: state.index should always be last element for children, right? Why you don't return just newChild  there?

AlexDM0: Because the tabBar disappears when I do that. I've tested it with 3.2.5. The state that is provided to the method is that of the tabBar.

aksonov: But popToTop should not be called for TabBar, i believe (like pop, it doesn't have any sense)

AlexDM0: In my case I need it for the tab bar. The behaviour is based off the apple app store. Tapping the tab on which you already are is back to root of the tab. Is there a cleaner way to implement this? I could put it in the reducer of course but I thought it was a nice addition to the actions.

aksonov: Oh, i see now. Does it work for stack too?

aksonov: Another thought that even for tab bar all tabs are actually stacks (with name tab+'_'), so popToTop should be applied to them, not to tab bar...

aksonov: So could you make it in the same way as POP action ? Then it will work for both tabs/stacks and it will be much simpler (just return first single child as children array)

AlexDM0: I just worked with the incoming state in the update method.. I'm not sure what would change for stacks, though the assert I copied over from the jump action currently limits it to tabs. I can remove that? What would the expected behaviour be for the stacks? As long as there is a shown back button this action makes sense, though thats not always the case.

AlexDM0: The main issue I had here was that if I used reset, the tabBar would only show one of my tabs. For stacks I assume the reset Action would work?

aksonov: Have you tried Action.callback({key:NAME+'_', type:'reset'}) ?

AlexDM0: Wouldnt that mean I'd have to hardcode it for each of my tabs? I have not named my tabs as in the example, seemed arbitrary? Is the naming convention important here?

This is my router code with a small set of views in there.
```js
 return <Router createReducer={reducerCreate} store={store}>
      <Scene key="tabBar" tabs={true} tabBarStyle={{backgroundColor:colors.menuBackground.h}}>
        <Scene key="overview" title="Overview" icon={TabIcon} iconString="ios-color-filter-outline" navigationBarStyle={{backgroundColor:colors.menuBackground.h}} titleStyle={{color:'white'}} renderBackButton={backButtonFunction}>
          <Scene key="groupOverview" component={GroupOverview} title="Group Overview" />
          <Scene key="roomOverview"  component={RoomOverview} onRight={onRightFunctionEdit} rightTitle="Edit" rightButtonTextStyle={{color:'white'}} />
          <Scene key="roomEdit"      component={RoomEdit} title="Edit Room" />
          <Scene key="deviceEdit"    component={DeviceEdit} title="Edit Device" />
        </Scene>
        <Scene key="settings" title="Settings" icon={TabIcon} iconString="ios-gear-outline" navigationBarStyle={{backgroundColor:colors.menuBackground.h}} titleStyle={{color:'white'}} renderBackButton={backButtonFunction}>
          <Scene key="Settings" component={SettingsOverview} title="Settings"/>
        </Scene>
      </Scene>
    </Router>;
```

Whenever I refer to specific scenes inside the tabs, the tabbar disappears.

If I have to hardcode the action key in the reducer for each of my tabs it seems a bit more sloppy than using the index of the state?

aksonov: So if you set type='reset' for groupOverview and then call Actions.groupOverview() when user clicked on tab, will it work?

AlexDM0: It works dynamically if I change the reducer like this:
```js
const reducerCreate = params=> {
  return (state, action)=>{
    if (state) {
      let currentTabIndex = state.index;
      if (state.children[currentTabIndex].name === action.key && action.type === 'jump') {
        return defaultReducer(state, {key: state.children[currentTabIndex][0].name, type:'reset'})
      }
    }
  }
};
```

I guess thats less code than my proposal :)

aksonov: Maybe you could propose PR based on it? For cases when first child has type 'reset'...

AlexDM0: I'm not sure how that would look for a general case.. I have no type reset on the child, its being forced into the reducer at the moment. It could be the default behaviour of the tabs since it's also done that way in iOS while Android doesnt use tabs..

rgovindji:
@AlexDM0 I tried that solution but it doesn't work 100%. I have to click on the tab icon twice in order for the page to show. I'm not really good with reducers, do I need to kick off two actions? One for the parent tab and one for the child?

AlexDM0:
Hi, 

I explicitly wanted it to be done at the second tap. The first tap is just to switch. 

Regards

rgovindji:
Gotcha, instead i just reset the child component when I navigated away to another page. Thanks for the help!

PR #479, opened 2016-04-07
Remove peerDependencies

Mokto:
See https://github.com/oblador/react-native-vector-icons/issues/76#issuecomment-165959265

Unable to install RCs.


bosung90:
+1

Issue #489, opened 2016-04-10
tracking React Native 0.24 compatibility/issues


joenoon:
I'm opening this to track problems and gather information as I dig into this.  If anyone has ideas or figures it out first, please update here.  It looks like after upgrading to RN 0.24 there is an issue in DefaultRenderer not receiving the expected navigationState, and getting:

```
undefined is not an object (evaluating 'navigationState.index')
```

Logging the keys of navigationState inside DefaultRenderer, I see:
```
 [ 'layout',
  'navigationState',
  'onNavigate',
  'position',
  'scene',
  'scenes' ]
```

So it seems like its not getting the right value here for navigationState in 0.24.  I did scan through the diff of NavigationExperimental on react-native between 0.22 and 0.24 but didn't see anything obvious.

I'll update here if I figure it out.

joenoon:
I think I found a fix in the PR above

sylvainbaronnet:
I upgraded RNRF to 3.22.22 and now the custom NavBar doesn't show up at all.

With RN 0.24.1

sylvainbaronnet:
I tried previous versions of RNRF and the last version where the custom NavBar show up is the 3.22.8
With the 3.22.11 there's the bug. There's a change between these version that introduce the bug.

(3.22.9 & 3.22.10 was not published to npm so I can't try them)

joenoon:
In case other people land here, as of today I'm using this combination successfully:

```
    "react-native": "0.24.1",
    "react-native-router-flux": "aksonov/react-native-router-flux#3.24"
```

sylvainbaronnet:
There's still the problem with `type="replace"` but it's a RN bug

jamesone:
@sylvainbaronnet  Any update on a fix for this?

aksonov:
@joenoon Could you check if it is possible to merge 3.24 with latest branch and upgrade it to usage with 0.26-rc? This way we will create new branch 3.26 and then will replace master when official 0.26 will be released.

sylvainbaronnet:
@aksonov import must be updated for 0.25+, there's a warning currently. I can submit a PR if you want (to 3.24 ? wait 3.26 ?)

aksonov:
@sylvainbaronnet Maybe could you integrate 3.24 into master and then update imports?

sylvainbaronnet:
Sorry I deleted my message because I realized imports were already updated in master.

sylvainbaronnet:
[...]

PR #490, opened 2016-04-10
change TabBar renderScene per RN 0.24 signature change

joenoon:
was (route,index), now NavigationSceneRendererProps object.

should maintain 0.22 compatibility.  can be cleaned up once 0.22 compatibility isnt needed.

https://github.com/aksonov/react-native-router-flux/issues/489

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ class App extends React.Component {
 | type | string | 'push' or 'jump' | Defines how the new screen is added to the navigator stack. One of `push`, `jump`, `replace`, `reset`. If parent container is tabbar (tabs=true), jump will be automatically set.
 | tabs| bool | false | Defines 'TabBar' scene container, so child scenes will be displayed as 'tabs'. If no `component` is defined, built-in `TabBar` is used as renderer. All child scenes are wrapped into own navbar.
 | initial | bool | false | Set to `true` if this is the initial scene |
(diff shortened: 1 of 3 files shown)


aksonov:
Looks great, thanks, but could we leave 'duration' compatibility? I.e. pass applyAnimation by Default Renderer if duration is defined (and version is 0.24+)? Also about direction...

joenoon:
@aksonov updated.  I put direction back, but I'm not sure if it works any more with RN upstream.  Just glancing through, I see mentions of direction in NavigationCardStack.  So it possibly moved.  But better if someone who knows what the expected behavior feels like can check it out.

aksonov:
Looks like direction doesn't work #499. I just thought you know how to implement it via new methods like you did with 'duration'

joenoon:
@aksonov I think its a bit more complicated.  From what I can tell so far, direction moved into NavigationCardStack.

I started experimenting with using NavigationCardStack in DefaultRenderer, but I'm not sure thats the answer because then we lose things like custom animations completely (what used to be setTiming, now applyAnimation).

To support direction outside of NavigationCardStack looks like maintaining a copy/paste derivative of it, which is likely to break or be hard to keep track of given how much is still changing with NavigationExperimental.

So I'm just not really sure what the right approach is.

PR #503, opened 2016-04-11
direction support for rn 0.24+

joenoon:
https://github.com/aksonov/react-native-router-flux/pull/490

This breaks < 0.24.  I thought about trying to make it backward compatible, but seems like it would get messy and, at best, not crash.  What are the compatibility expectations?

### DIFF
--- a/src/DefaultRenderer.js
+++ b/src/DefaultRenderer.js
@@ -64,7 +64,11 @@ export default class DefaultRenderer extends Component {
             if (duration === null || duration === undefined) duration = navigationState.duration;
             if (duration !== null && duration !== undefined) {
                 optionals.applyAnimation = function (pos, navState) {
-                    Animated.timing(pos, {toValue: navState.index, duration}).start();
+                    if (duration === 0) {
+                        pos.setValue(navState.index);
+                    } else {
+                        Animated.timing(pos, {toValue: navState.index, duration}).start();
+                    }
                 };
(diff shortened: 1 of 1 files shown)


sylvainbaronnet:
I guess this module can require RN 0.24 (when it's released) for next releases ? It's better than not supporting latest RN version...

aksonov:
Great work! 0.24 is not released officially yet, so i will merge it once it is out (next Monday?), ok?

aksonov:
Another possibility is to publish 'rc' release for the component which will work with 0.24.rc

philipshurpik:
Thanks!

But this PR has an issue:
disabling panHandlers with panHandlers={null} stopped working

It will be better to do smth like that:
`var state = props.scene.navigationState;
 const panHandlers = state.panHandlers !== undefined ? state.panHandlers : (isVertical ?
                NavigationCardStackPanResponder.forVertical(props) :
                NavigationCardStackPanResponder.forHorizontal(props));`

aksonov:
Close it because of #536 (that PR includes this one)

yusufseyrek:
RN 24.0 is released but direction still not working

aksonov:
You may try 3.24.0 version which supports RN 0.24. However you could check #557

yusufseyrek:
Thanks a lot @aksonov you are the best!

PR #509, opened 2016-04-12
Add getInitialState API to Scene and Router

lelandrichardson:
This PR adds a `getInitialState` prop that can be passed into the `<Scene />` or `<Router />` components  which will be run each time an instance of the scene is added to the stack. This is a use case not supported by the current props being added to the navigation state, which will only allow for static constants being passed to the scene. With `getInitialState`, you can create state that persists for the entirety of the scene, but is unique across instances of the same scene. This is useful for a variety of cases.

The API is done in such a way that you can define a `getInitialState` function for a parent scene and it will get passed to a child scene, but it will be merged in with the `getInitialState` calls of the children scenes.

### DIFF
--- a/src/State.js
+++ b/src/State.js
@@ -8,10 +8,34 @@
  */
 import assert from "assert";
 
+function getStateFromScenes(route, scenes, props) {
+    const getters = [];
+    let result = {};
+    let scene = route;
+    while (scene) {
+        if (scene.getInitialState) {
+            getters.push(scene.getInitialState);
+        }
+        scene = scenes[scene.parent];
+    }
+
+    getters.reverse().forEach(fn => {
+        result = {...result, ...fn(props)};
+    });
+
+    return result;
+}
+
 export function getInitialState(route:{string: any},scenes:{string:any}, position=0, props={}){
     const {key, style, type, ...parentProps} = props;
     if (!route.children){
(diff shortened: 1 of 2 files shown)


lelandrichardson:
Note: This PR includes the commit from #504 and will go away when/if that is merged.

aksonov:
Strange, i see 'this branch has conflicts', could you check?

lelandrichardson:
@aksonov There was a conflict with some tests added in a different commit. Should be fine now.

Thanks for merging these in! Let me know if you have any questions. Thank you for your great work on this library - if we can get these changes in then we should be able to use it at Airbnb!

aksonov:
@lelandrichardson Thanks for PRs! Currently I have some troubles (no troubles actually but lack of time) to migrate to 0.23-0.24rc (seems no sense to migrate to 0.23, but better to migrate to 0.24), are you using 0.22?

lelandrichardson:
@aksonov we are currently on 0.22 but going to upgrade soon. Probably when 0.24 comes out.

I can take a stab at putting a PR out to upgrade this library once we do, if you want?  Looks like #503 is already moving in the right direction.

BTW, when can I expect these merged PRs to be published to npm?  Thanks.

aksonov:
Yes, #503 and probably some other will do it. 3.2.11 is already published, please check.

lelandrichardson:
Thanks for such a quick turnaround!

PR #516, opened 2016-04-12
Allow getInitialState to be passed into Router as well as Scene

lelandrichardson:
This was a missing piece of #509 that allows a `getInitialState` function to be supplied to Router, which will then be applied to every child scene under it.

Let me know if there are any questions!

### DIFF
--- a/src/State.js
+++ b/src/State.js
@@ -19,6 +19,10 @@ function getStateFromScenes(route, scenes, props) {
         scene = scenes[scene.parent];
     }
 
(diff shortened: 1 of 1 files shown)


aksonov:
Maybe it is redundant, since all properties all of Router will be passed to all scenes anyway, right?

lelandrichardson:
@aksonov interesting. That's kind of what I thought, too... but it didn't work for me when I tried... :/

PR #530, opened 2016-04-15
add new feature about title

timzaak:
1.make the title delayInit
2.expose callback of the press of title.


aksonov:
Looks nice, could you add also modification to README?

timzaak:
I think #547 would be more nicer. Because this pr can not change the title dynamic, and #547 can also render images on the title position. :)

PR #531, opened 2016-04-15
Add FOCUS action, dispatch when react-redux is available

lynndylanhurley:
### Problem:

My app needs to know which screen is currently in focus. Because the `*_FOCUS` actions were removed in v3, this is no longer possible.

### Solution:

Implement the `FOCUS_ACTION` action. Dispatch the event when redux-router is present. The action payload contains the props for the screen that is currently in focus (`name`, `title`, etc.).

Usage is as follows:

~~~js
// reducers/routes.js

import { createReducer } from 'redux-immutablejs';
import Immutable from 'immutable';
import {Actions} from 'react-native-router-flux';

const {FOCUS_ACTION} = Actions;

const initialState = Immutable.fromJS({
  scene: {},
});

export default createReducer(initialState, {
  [FOCUS_ACTION]: (state, {scene}) => state.merge({scene}),
});
~~~

##### EDIT:

Instead of importing `react-redux`, it's better to pass the `dispatch` method to the router. This frees RNRF from any dependencies.

So usage from the router should be as follows:

~~~js
// routes.js

import React, {PropTypes} from 'react';
import {
  Actions,
  Router,
  Scene,
} from 'react-native-router-flux';
import Dashboard from './scenes/Dashboard';

const scenes = Actions.create(
  <Scene key="Root">
    <Scene key="Main">
      <Scene
        key="Dashboard"
        title="Dashboard"
        component={Dashboard}/>
    </Scene>

    {/* ... more scenes */ }
  </Scene>
);

class Routes extends React.Component {
  static propTypes = {
    dispatch: PropTypes.func,
  };

  render () {
    return (
      <Router
        dispatch={this.props.dispatch} <-- pass the dispatch method to the router
        scenes={scenes} />
      );
  }
}

// make sure to "connect" the routes component
export default connect()(Routes);
~~~

That's it!

### DIFF
--- a/src/Actions.js
+++ b/src/Actions.js
@@ -44,7 +44,7 @@ class Actions {
         this.init = this.init.bind(this);
         this.pop = this.pop.bind(this);
         this.refresh = this.refresh.bind(this);
-
+        this.FOCUS_ACTION = FOCUS_ACTION;
     }
 
     iterate(root: Scene, parentProps = {}, refs = {}) {
@@ -111,6 +111,10 @@ class Actions {
         let refs = {};
         this.iterate(scene, {}, refs);
         return refs
+    }
+
+    focus(scene){
+        return {type: FOCUS_ACTION, scene};
     }
 }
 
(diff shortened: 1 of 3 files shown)


caledhwa:
+1 looks like a good solution!!!

lynndylanhurley:
Thanks @caledhwa!!

@aksonov - what do you think about this PR?

lynndylanhurley:
See also #537.

aksonov:
Close it in favour for #537

PR #533, opened 2016-04-16
add more details about tabbar

timzaak:
1. remove the useless default='...'
2. add more details about tabbar.
this cost me lots of time to figure out why the navigation bar do not act as I expected. Hope to help others :)


README.md
```diff
@@ -253,6 +253,8 @@ const reducerCreate = params=>{
 return <Router scenes={scenes} createReducer={reducerCreate} />;
 
 ```
+## Tabbar
+Every tab has its own navigation bar. However, if you do not set its parent `<Scene tabs={true} />` with `hideNavBar={true}`, the tabs' navigation bar will be overrided by their parient.
```
sylvainbaronnet: parient -> parent

PR #537, opened 2016-04-17
Implement FOCUS_ACTION using createReducer approach

lynndylanhurley:
This is an alternative approach to #531 that follows @aksonov's intended implementation.

### Problem

My app needs to know which screen is currently in focus. Because the *_FOCUS actions were removed in v3, this is no longer possible.

### Solution

Implement the FOCUS_ACTION action. The action payload contains the props for the screen that is currently in focus (name, title, etc.).

### Usage

~~~js
// actions/routes.js

export const FOCUS_ACTION = 'FOCUS_ACTION';

export function focus(scene) {
  return {type: FOCUS_ACTION, scene};
}
~~~

~~~js
// reducers/routes.js
import { createReducer } from 'redux-immutablejs';
import Immutable from 'immutable';
import {FOCUS_ACTION} from '../actions/routes';

const initialState = Immutable.fromJS({
  scene: {},
});

export default createReducer(initialState, {
  [FOCUS_ACTION]: (state, {scene}) => state.merge({scene}),
});
~~~

~~~js
// routes.js

import {focus} from './actions/routes';
// other imports...

class Routes extends React.Component {
  static propTypes = {
    dispatch: PropTypes.func,
  };

  reducerCreate(params) {
    return (state, action) => {
      if (action.type === 'focus') {
        this.props.dispatch(focus(action.scene))
      }
    };
  }

  render () {
    return (
      <Router
        createReducer={this.reducerCreate.bind(this)}
        scenes={scenes} />
    );
  }
}

export default connect()(Routes);
~~~

~~~js
// app.js

import Routes from './routes';
// other imports...

// create store...

class App extends React.Component {
  render () {
    return (
      <Provider store={store}>
        <Routes />
      </Provider>
    );
  }
}
~~~

### DIFF
--- a/src/Actions.js
+++ b/src/Actions.js
@@ -44,14 +44,14 @@ class Actions {
         this.init = this.init.bind(this);
         this.pop = this.pop.bind(this);
         this.refresh = this.refresh.bind(this);
-
+        this.focus = this.focus.bind(this);
     }
 
     iterate(root: Scene, parentProps = {}, refs = {}) {
         assert(root.props, "props should be defined for stack");
         const key = root.key || "root";
         assert(key, "unique key should be defined ",root);
-        assert([POP_ACTION, POP_ACTION2, REFRESH_ACTION, REPLACE_ACTION, JUMP_ACTION, PUSH_ACTION, RESET_ACTION, "create",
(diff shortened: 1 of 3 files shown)


aksonov:
It would be great if you add it to README.

lynndylanhurley:
@aksonov - I will do that. Thanks for the merge :smiley:

Issue #460, opened 2016-04-04
How to change back button ? Any examples ?


melihmucuk:
I can't change default back button on iOS. Whats the problem ? I tried `renderBackButton`, `renderLeftButton` and `backButtonImage` props, but didn't change.

```
<Scene
  key="entry"
  component={Entry}
  title="Title"
  titleStyle={styles.title}
  navigationBarStyle={styles.navBar}
  renderLeftButton={() => this._backButton.bind(this)}
  renderBackButton={() => this._backButton.bind(this)}
  backButtonImage={backButton}/>
```

`var backButton = (<Image source={this.state.backIcon} style={{width: 30, height: 30}} />);`

```
_backButton(){
    return (
      <View style={{paddingHorizontal: 5, paddingVertical: 7}}>
        <TouchableOpacity
          style={{padding: 10}}
          onPress={() => Actions.pop()}>
          <Icon
            name="arrow-left"
            size={20}
            color="#e67e22" />
        </TouchableOpacity>
      </View>);
  }
```

### Version
```
"dependencies": {
    "react": "^0.14.8",
    "react-native": "^0.22.2",
    "react-native-router-flux": "^3.2.6",
    "react-native-vector-icons": "^1.3.3"
  }
```

SudoPlz:
I could not get it to work either, till then I just go and tamper the `back_chevron.png` img in the `node_modules/react-native-router-flux/src` directory.

lee-gray:
@melihmucuk try adding it to your root Scene, I don't know why but it worked for me.

<Scene key="root" backButtonImage={require('../images/back_chevron.png')}>
    <Scene key="foo" component={fooView} title="Foo" />
    <Scene key="bar" component={barView} title="Bar" />
</Scene>

AlexDM0:
I just used the renderBackButton method:

renderBackButton={backButtonFunction}

with

```js
let backButtonFunction = function() {
  if (this.props.navigationState.index === 0) {
    return null;
  }
  return (
      <TouchableOpacity style={[{
        width: 100,
        height: 37,
        position: 'absolute',
        bottom: 4,
        left: 2,
        padding: 8,
        justifyContent:'center',
    }]} onPress={Actions.pop}>
        <View style={{flexDirection:'row', alignItems:'center'}}>
          <Icon name="ios-arrow-back" size={25} color={'#ffffff'} style={{marginTop:2,paddingRight:6}} />
[...]

Issue #478, opened 2016-04-07
Seeking Guidance on Notification Icon Implementation


ringzhz:
Hey,

  Great work! This library is the only one I could get working well for Navigation w/ Toolbar integration on react-native android.
I'm adding a custom notification icon to the notification bar by using this snippet: 
`<Scene key="root" renderRightButton={()=>notificationsIcon}>`

I was wondering if you have any guidance on passing in a state to this icon so that it might be able to conditionally render itself as having no notifications or n-number of notifications.

Versions:
    "react-native": "^0.23.0"
    "react-native-router-flux": "^3.2.7"

ringzhz:
Additionally, I'm trying to use a custom icon for the leftButton.

I'm not sure how to write comparable functionality in my button renderers without a reference to the navigationState.

I can understand why you haven't done it yet, but might it be useful to bind the passed in renderers to `this`, as is done for the built-in renderers. How else would my button be able to access the drawer instance as to call its `toggle` method.

Unless I'm missing something, which is the likely case, I'd suggest the following change. I will gladly create a PR if you agree.

Rewriting this:
```
export default class NavBar extends React.Component {
    componentWillMount(){
        const state = this.props.navigationState;
        this._renderRightButton = this._renderRightButton.bind(this);
        this._renderBackButton = this._renderBackButton.bind(this);
        this._renderLeftButton = this._renderLeftButton.bind(this);
        this._renderTitle = this._renderTitle.bind(this);
        if (state.renderLeftButton){
            this._renderLeftButton = state.renderLeftButton;
        }
        if (state.renderRightButton){
            this._renderRightButton = state.renderRightButton;
        }
        if (state.renderBackButton){
            this._renderBackButton = state.renderBackButton;
        }
    }
```

as:
```
export default class NavBar extends React.Component {
    componentWillMount(){
        const state = this.props.navigationState;
        this._renderRightButton = this._renderRightButton.bind(this);
        this._renderBackButton = this._renderBackButton.bind(this);
        this._renderLeftButton = this._renderLeftButton.bind(this);
        this._renderTitle = this._renderTitle.bind(this);
        if (state.renderLeftButton){
            this._renderLeftButton = state.renderLeftButton.bind(this);
            //or .bind(this.props.navigationState);
        }
        if (state.renderRightButton){
            this._renderRightButton = state.renderRightButton.bind(this);
        }
        if (state.renderBackButton){
            this._renderBackButton = state.renderBackButton.bind(this);
[...]

Issue #213, opened 2016-02-15
renderRightButton with save button.


fionita:
How to pass data from component form, to the Route.

SudoPlz:
Did you solve this? How ? Could you please share? Thank you.

fionita:
I've hided the router flux navbar and used toolbarAndroid instead.

SudoPlz:
@fionita didn't that result in messy code? If I understand correctly you no longer use the ExperimentalNavigator that comes with the navbar, but you're using `ToolbarAndroid` with Android and `NavigatoriOS` with iOS ? 
So you actually manage the `ToolbarAndroid` and `NavigatoriOS` on each schene seperately?

PR #545, opened 2016-04-18
renderRightButton and renderLeftButton takes selected component as argument

jondot:
Including the selected child lets the renderRight and Left buttons be able to render based on the child's state.

```javascript
  renderRightButton={(props)=>{ <RightButton item={props.item}/> } }
```
This can help with item counts, save/create state for items that are
existing vs new, and more.

Right now this uses 'selected' which includes the entire component,
it would probably be more suitable to include just the props of that
component but I'm not sure that's possible at that point in time.

Additionally, this should provide a way to solve:

* https://github.com/aksonov/react-native-router-flux/issues/213
* https://github.com/aksonov/react-native-router-flux/issues/478

### DIFF
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -65,7 +65,7 @@ export default class NavBar extends React.Component {
             <Animated.View
                 style={[styles.header, state.navigationBarStyle, selected.navigationBarStyle]}>
                 {state.children.map(this._renderTitle, this)}
-                {renderLeftButton(selected) || renderBackButton()}
+                {renderLeftButton(selected) || renderBackButton(selected)}
                 {renderRightButton(selected)}
             </Animated.View>
         );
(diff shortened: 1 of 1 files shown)


AlexDM0:
Could we also do insert this in the renderBackButton to fix #460 ? I wanted to make a pull myself but if you're doing it anyway might as well add the selected to the renderBackButton here :)

jondot:
Sure, I've read #460, would `renderBackButton(selected)` help?

AlexDM0:
Yes. Granted the methods need to be adapted but we'd have what we'd need I think.

jondot:
Should be good now.
This solution in general covers some of the cases, but for some, it's not enough sadly.
It might be worth to watch this https://github.com/ericvicenti/navigation-rfc/issues/70 meanwhile.

PR #550, opened 2016-04-19
Updated Actions values to a more standard format

mmazzarolo:
Is there any reason for using lowercase actions?  This actions name are the ones used by Redux/Flux and 99% of the time the actions used with redux are uppercase :)
I and also added `"ROUTER_"` in front of them.  
Let me know what you think of the change (I'm testing it only since two hours ago and it seems to not break the router).
**It is a breaking change for anyone using the previous actions names with redux/flux**

P.S.: I can also update the docs if needed


aksonov:
POP_ACTIONs should have exactly the same names used in NavigationExperimental because of swipe back behaviour. The same for probably some other actions...

mmazzarolo:
Ops, didn't know about it. Thanks.

Issue #561, opened 2016-04-20
How to get the navigation bar Dimensions from within a Schene?


SudoPlz:
For example lets say I'm within my Login schene. How can I access the navigation bar and get for example its height?

Thank you..!

aksonov:
Sorry, it is out of scope of this component, it is about react native programming. I guess you should use onLayout handler

SudoPlz:
Could this perhaps be a future feature or should I just close the issue?

aksonov:
It is not planned, but you may submit PR.

SudoPlz:
On which file should I look for the Navigation bar component and its styles?

aksonov:
NavBar.js

SudoPlz:
Ok thanks!

SudoPlz:
Hey @aksonov , I found a way to get the dimensions of my NavBar, but I'm currently trying to figure out a way to pass that object with the width and height back to the user. If your project was using redux, I would dispatch an action to save it to the current state, but I'm so confused on how to pass that object to the end user with your library. Any suggestions please?

aksonov:
Use Actions.refresh(data). You may specify key of scene where you want to pass that data.

SudoPlz:
Ok so since my pull request was not accepted, here is how to set your View to render below the nav bar in both iOS and Android: 

`paddingTop:Platform.OS === 'ios' || Platform.Version > 19 ? 64 : 44,  //nav bar height`

af7:
@SudoPlz Is this a cross-device solution?
I find that on my Samsng Android I need to use "53", so how can we be sure of the values?

SudoPlz:
@af7 The code is no more valid. I got this code from https://github.com/aksonov/react-native-router-flux/blob/0.26-stable/src/NavBar.js#L58 but apparently this has changed in the latest version. 

Now the code has changed to :

```js

af7:
@SudoPlz Thanks, may I assume that this is regardless of pixelRatio and same for all models of Android and iOs?

SudoPlz:
@af7 Yes!

PR #566, opened 2016-04-21
Added a way to know the nav bar dimensions using the refresh action.

SudoPlz:
I'm creating this pull request to add the feature I requested in : https://github.com/aksonov/react-native-router-flux/issues/561

### DIFF
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -63,6 +63,10 @@ export default class NavBar extends React.Component {
         let renderBackButton = selected.renderBackButton || this._renderBackButton;
         return (
             <Animated.View
+                onLayout={(event) => {
+                    var {width, height} = event.nativeEvent.layout;
+                    Actions.refresh({navBarDimensions:{width:width, height:height}});
+                }}
                 style={[styles.header, state.navigationBarStyle, selected.navigationBarStyle]}>
                 {state.children.map(this._renderTitle, this)}
                 {renderBackButton() || renderLeftButton()}


SudoPlz:
Ok check this out and let me know if it is correct, I can now get the dimension of the nav bar in my reducer like so:

    const reducerCreate = params=>{
                return (state, action)=>{

                  switch(action.type){
                    case "refresh":
                      console.log("Nav bar dimensions:",action.navBarDimensions.height); 
                      break;
                    default:
                      break;
                  }
                }
            };

aksonov:
Why you need custom reducer? Current scene will get navBarDimensions property.

SudoPlz:
Hmm sure but I'm new to react-native-router-flux, how would I access it otherwise?

aksonov:
Oh, if you need to access it from your reducer, not from your scene component, then you did right.

rturk:
Why do you need this? I don't know about your app but storing dimensions in redux doesn't seems to be best practice.

aksonov:
agree with @rturk

SudoPlz:
Well then what should I do, would `this.props. navBarDimensions` work in any Schene now?

aksonov:
Good question, i'm not sure. It will 100% work for current scene when layout is changed, but for other scenes it will be lost probably. Let me think about it.

SudoPlz:
`this.props.navBarDimensions` is nowhere to be found at the moment (answering my own question above). null null null

aksonov:
Maybe storing nav bar dimensions in redux is only way...
Here is one solution:
1. don't use 'refresh' but your redux dispatch
2. connect RNRF Router to your redux.
3. all properties you passed to Router from 'connect', will be globally available across all scenes

rturk:
@SudoPlz can you detail your use case, why you need dimensions?

We can explore it at Gitter. I have a fairly complex App and I've managed to make it work only using flexbox, padding, and margin in the root View of each component

Overall IMO this package is a navigation stack. Adding rendering, view, features will diverge from it's core purpose

aksonov:
@rturk I don't have different height for navbar, but probably will do (smaller height for landscape, for example). One usage i see now is setting top:HEIGHT within my scenes to avoid its overlapping by navbar. 
In my current app it is just fixed (equal to navbar height), but what i can do it navbar height is flexible?

SudoPlz:
Ok this comment is so confusing to me right now. 

This is pretty much my `init.js` file at the moment:

      /**
      *
      * ## Initial state
      * Create instances for the keys of each structure in placeAVote
      * @returns {Object} object with 4 keys
      */
      function getInitialState() {
        const _initState = {
          auth: new authInitialState,
          device: (new deviceInitialState).set('isMobile',true),
          global: (new globalInitialState),
          profile: new profileInitialState,
          router: new routerInitialState
        };
        return _initState;
      }


let Example = React.createClass( {
          render() {


            const store = configureStore(getInitialState());

            //Connect w/ the Router
            const Router = connect()(RNRF.Router);

            // configureStore will combine reducers from Example and main application
            // it will then create the store based on aggregate state from all reducers


            var self = this;
            var renderRtBtn = function renderRtBtn(){
              return
            }
            const reducerCreate = params=>{
                return (state, action)=>{

                  switch(action.type){
                    case "push":
                      store.dispatch(navigateState(action.key));    //I navigate to the state I want to go using my action creator (uses Actions.ScheneName on the background)
                      break;
                    case "back":
                    case "BackAction":
                      store.dispatch(navigateToPreviousState());    //I navigate to the previous state using my action creator (uses Actions.pop on the background)
                      break;
                    case "refresh":
                      console.log("Nav bar height:", action.navBarDimensions.height); 
                      store.dispatch(setNavBarDimensions(action.navBarDimensions));
                      break;
                    default:
                      break;
                  }
                }
            };

            // let backButtonImg = ()=><Image source={require("../assets/back_chevron.png")} style={{padding: 8,backgroundColor:'red'}}></Image>;
            //
            // <Scene key={ScheneKeys.LOGIN} type="push" title="Sign In" navigationBarStyle={{backgroundColor:Colors.primaryColor}} backButtonImage={backButtonImg} renderRightButton={()=><RightPavLogo/>}>
            // setup the router table with App selected as the initial component

            // when the hideNavBar is true we see the right button but not the left, and the opposite.
            return (
              <Provider store={store}>
                <Router hideNavBar={false} createReducer={reducerCreate} sceneStyle={{backgroundColor:'#F7F7F7'}} >
                    <Scene key="root" navigationBarStyle={{backgroundColor:Colors.primaryColor}}  titleStyle={{color:Colors.mainTextColor, fontFamily:"Whitney"}}>
                      <Scene key={ScheneKeys.ONBOARDING} direction="vertical" component={Onboarding} type="replace" hideNavBar={true} />
                      <Scene key={ScheneKeys.LOGIN} component={EmailSignIn} hideNavBar={false} title="Sign In" type="push" renderBackButton={()=><BackBtnImg/>}  renderRightButton={()=><RightPavLogo/>}/>
                      <Scene key={ScheneKeys.REGISTER_STEP_1} component={EmailSignUpStep1} type="push" hideNavBar={true} />
                      <Scene key={ScheneKeys.REGISTER_STEP_2} component={EmailSignUpStep2} type="push" hideNavBar={true} />
                      <Scene key={ScheneKeys.REGISTER_STEP_3} component={EmailSignUpStep3} type="push" hideNavBar={true} />
                      <Scene key={ScheneKeys.REGISTER_STEP_4} component={EmailSignUpStep4} type="push" hideNavBar={true} />
                      <Scene key={ScheneKeys.TOPIC_PICK} component={TopicPick} schema="modal" type="push" hideNavBar={true} />
                      <Scene direction="vertical" key={ScheneKeys.NEWSFEED} title="News Feed" hideNavBar={true} panHandlers={null} duration={1} tabs={true} initial={true}>
                            <Scene key={ScheneKeys.TAB_NEWS} title="News Feed" component={NewsFeed} icon={TabIconFactory} navigationBarStyle={{backgroundColor:Colors.primaryColor}}  titleStyle={{color:Colors.mainTextColor, fontFamily:"Whitney"}}/>
                            <Scene key={ScheneKeys.TAB_NOTIFS} title="Notifications" component={NewsFeed} icon={TabIconFactory} navigationBarStyle={{backgroundColor:Colors.primaryColor}}  titleStyle={{color:Colors.mainTextColor, fontFamily:"Whitney"}} notifications={["a notification", "another notification"]}/>
                            <Scene key={ScheneKeys.TAB_PROFILE} title="Profile" component={Profile} icon={TabIconFactory} navigationBarStyle={{backgroundColor:Colors.primaryColor}}  titleStyle={{color:Colors.mainTextColor, fontFamily:"Whitney"}} initial={true}/>
                      </Scene>
                    </Scene>
                  </Scene>
                </Router>
              </Provider>
            );
          }
        });
        /**
        * registerComponent to the AppRegistery and off we go....
        */

        AppRegistry.registerComponent('ExampleApp', () => Example);

@rturk The reason I need the dimension of the navigation bar is because I want to know its height, so that I can start rendering my schene children below it. If otherwise some of my children will be hidden beneath the navigation bar. I have to specifically set my View to start at `top= navigationBar.height`.

@aksonov you said:
1. don't use 'refresh' but your redux dispatch
But the Actions.Refresh() call is currently within the `react-native-router-flux` code, so I cannot use my own redux dispatch within there can I?

2. connect RNRF Router to your redux.
I think I have that in my code in line: `const Router = connect()(RNRF.Router);` but I neeever really understood how it works. I NEVER managed to get dispatches from RNRF directly within my own reducers.

3. all properties you passed to Router from 'connect', will be globally available across all scenes
Would you please mind giving a line of code or 2 as an example to this statement?

aksonov:
Maybe @rturk has better solution for it, i would listen it too.

aksonov:
1. You may use your own NavBar, and dispatch redux there, right?
2. connect(state=>({navBarHeight:state.navbar.navBarHeight})(RNRF.Router) will pass navBarHeight prop to all your scenes (but you should store it within your navbar reducer first)

rturk:
@aksonov @SudoPlz 
I'm using my own NavBar so have more control over he UI. Also trying to follow redux best practices, this means that I only store app data and app state in the redux store. 

Within my ownNavBar I'm listening to app state changes. If the app rotates I simply update the internal state of the NavBar height. Currently I only have two heights one for portrait one for landscape.

Also I wouldn't recommend using `top` for layout. Instead I'm using a View with a `padding` as wrapper for my content. React Native Virtual DOM is smart and will convert this to a single Native Object. The benefit of this is that different than top this will look much better for views that are rotated 

<View style={padding:navBarHeight}>
 ...content
</View>

SudoPlz:
@rturk Teach me sensei! Could I please see some code with this. Trying to style views using flexbox is already hard enough for me, would you be kind enough to share some code so that I can see your custom NavBar and how you implement this?

rturk:
@SudoPlz I'm OOTO with no access to my code. I'll share something mid next week.

sarovin:
I have the same prolem because the height of NavBar changes from IOS to Android and i want manage it in each screen.

SudoPlz:
So still no solution to this..

Issue #576, opened 2016-04-22
Customize <Tab>


evollu:
Do you think it is a good idea to add renderTab prop to allow custom Tab bar?
I can make a pull

cridenour:
You can already customize every component in the TabBar. 

```
<Scene key="main" tabs={true} component={TabBar}>
  <Scene key="showings" component={ShowingsList} title="Showings" icon={TabIcon} iconName="calendar" />
</Scene>
```

evollu:
I want to custom tabbar itself not tabIcon because I want to use TouchNativeFeedback instead of TouchOpacity in react-native-tab.

cridenour:
Please see the first line, you can pass a component to the scene with tabs={true}. I would suggest using the TabBar in this component as a start, and edit from there.

evollu:
yeah I tried and it works for most cases, but just not mine. Just want to add an option to not depend on another library.

cridenour:
Well, to be fair the tab library is the same maintainer as this library. 

But if you don't want to use it you won't need to. You have full control over that component as it just reads the navigation state to decide what to do. I personally overwrote mine to hide it when the user is logged out.

evollu:
I will change it in my fork then.
Thanks 😄

manideepcs024:
how can the tab text (which is key by default) can be changed or removed.

PR #582, opened 2016-04-24
add option to replace <Tabs> with custom tabbar

evollu:
if renderTabs is provided with tabs={true}, it will render the tab bar with custom component.
This provides more options to customization and solves my problem:
1. if I want to put tab bar at top rather than bottom
2. if I want to replace TouchableOpacity with TouchableNativeFeedback or my components


aksonov:
Sorry, why you can't just pass your TabBar instead? I don't want to make built-in TabBar more complex

evollu:
Sorry I may not get you correctly. It is not changing the built-in TabBar. If you don't have renderTabs prop then the built-in one will be rendered.

cridenour:
I thought we talked about this in #576 - you can use `component={MyTabBar}`

andreialecu:
Sorry for hijacking this. 

How would you use this to put the TabBar at the top? I tried something similar but had to move the order of the components around in `TabBar.js`. This kind of worked, except the TabBar now shows **above** the navigation bar.

evollu:
@cridenour you are right. passing component works. Sorry I didn't get your point in the other issue. Thanks a lot

PR #583, opened 2016-04-24
Adding an option to compute title for each screen

jondot:
This allows things like opening a detail view (in master-detail)
with the detail navbar title populated with the name of the item the user is
currently editing.

For example:

```
getTitle={(comp)=> comp.item ? comp.item.get('title') : "" }
```
Once the scene is navigated to, the navbar will use this specific
getTitle function, which will return a different title, based on
the navigation parameters that were used in order to navigate to it.


aksonov:
Sorry, it has conflicts now, please re-submit...

jondot:
Sure, should be good now.

PR #595, opened 2016-04-26
Actions code styles and tests fixed.

r0b1n:
Also `eslint` tests config added
I have removed all non-Action module testing with `getInitialState ` and `createReducer` because they are not relevant.


r0b1n:
@aksonov Please specify, does code assume what Actions.create should be used only once (Router mounted only once)?
Because in case of incorrect using i am getting on re-render:
```
Key launch is already defined!
Key home is already defined!
Key map is already defined!
// etc.
```

I suggest to add `isInitialized` variable and check if `Actions.create` was already invoked, then output message about double initializing.
What do you think?

```javascript
create(scene:Scene) {
    assert(scene, 'root scene should be defined');
    assert(!this.isInitialized,
      'Scene already initialized. ' +
      'Are you trying to render Router/Scenes more than once?'
    );
    const refs = {};
    this.iterate(scene, {}, refs);
    this.isInitialized = true;
    return refs;
  }
```

aksonov:
@r0b1n Good work! But why you removed reducer testing? It should be moved to other files, but not to be removed.
Maybe we should allow running Actions.create several times because of recent react native hot-reload feature .

r0b1n:
Yes, it should be moved to separate file, and it is broken because of `init` action removed in latest release.
Sure I an going to place it to separate test file.

PR #600, opened 2016-04-27
FIX TabBar.js ESlint

sarovin:


mikefowler:
👍

mikefowler:
All but `NavigationView` in the constant block at the top are unused; could probably remove the extraneous ones and re-indent there, as well.

```js
const {
  View: NavigationView,
} = NavigationExperimental;
```

PR #601, opened 2016-04-27
Fix eslint errors in Switch.js

mikefowler:
Also:

- Adds an eslint rule to allow console statements
- Fixes malformed JSON in `.babelrc`

### DIFF
--- a/src/Switch.js
+++ b/src/Switch.js
@@ -9,34 +9,41 @@ export default class extends Component {
     this.state = {};
   }
 
-  updateState(props) {
-    const navState = props.navigationState;
-    const selector = props.selector || console.error('selector should be defined');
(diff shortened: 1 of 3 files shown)


lelandrichardson:
LGTM.

src/Switch.js
```diff
@@ -9,34 +9,41 @@ export default class extends Component {
...
 
   componentWillReceiveProps(props) {
     this.updateState(props);
   }
+
+  updateState(props) {
+    const navState = props.navigationState;
+
+    const selector = props.selector;
+    if (!selector) console.error('Selector should be defined.');
```
r0b1n: Why not to stay in one style and use `assert` instead all these `console`? Then we may revert`.eslintrc` back.

PR #602, opened 2016-04-27
Cleanup Router.js

mikefowler:
Cleans up the eslint errors in `Router.js`. Previously, destructuring was being used to omit values from an object, which was creating unused reference errors. I've added lodash in order to use `omit` in place of the destructuring.

### DIFF
--- a/src/Router.js
+++ b/src/Router.js
@@ -6,63 +6,107 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-import React, { Component, NavigationExperimental } from 'react-native';
-const {
-    AnimatedView: NavigationAnimatedView,
-    Card: NavigationCard,
-    RootContainer: NavigationRootContainer,
-    Header: NavigationHeader,
-    } = NavigationExperimental;
+import React, {
+  Component,
+  NavigationExperimental,
+  PropTypes,
+} from 'react-native';
+
 import Actions from './Actions';
 import getInitialState from './State';
 import Reducer from './Reducer';
 import DefaultRenderer from './DefaultRenderer';
 import Scene from './Scene';
 
-export default class extends Component {
+const {
+  RootContainer: NavigationRootContainer,
+} = NavigationExperimental;
+
+const propTypes = {
+  dispatch: PropTypes.func,
+};
+
+class Router extends Component {
+
   constructor(props) {
     super(props);
     this.state = {};
-    this._renderNavigation = this._renderNavigation.bind(this);
-    this._handleProps = this._handleProps.bind(this);
-  }
-
-  _handleProps(props) {
-    const scenesMap = props.scenes || Actions.create(Array.isArray(props.children) ? <Scene key="__root" hideNav {...this.props}>{props.children}</Scene> : props.children);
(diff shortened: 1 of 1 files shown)


lelandrichardson:
This PR in general LGTM. This does add a dependency on `lodash/omit`, which I think is reasonable, but I want to allow someone else to comment on that as well.

mikefowler:
👍

For the group: an alternative is to disable the eslint rule for unused references on the line that uses destructuring to omit properties:

```js
// eslint-disable-next-line no-unused-vars
const { children, style, scenes, reducer, createReducer, ...parentProps } = props;
```

cridenour:
Why not try and mirror react-native's lodash dependency to avoid any actual additional requirements?

package.json
```diff
@@ -36,6 +36,7 @@
   ],
   "dependencies": {
     "assert": "^1.3.0",
+    "lodash": "^4.11.1",
```
r0b1n: Does whole `lodash` necessary because of one function?

r0b1n:
I suggest to have own stack of needed functions to avoid dependencies.
`omit`, `pick`, `compare`

Move them to helpers/utils file and reuse in components.

lelandrichardson:
@mikefowler maybe we just use the proposed ignore comment + object-rest?

```
// eslint-disable-next-line no-unused-vars
const { children, style, scenes, reducer, createReducer, ...parentProps } = props;
```

This way the behavior is sure to be the same, and we have no question of additional dependencies?

charpeni:
@mikefowler @lelandrichardson 

I'm curious about this declaration :
```javascript
static propTypes = {
  dispatch: PropTypes.func,
};
```

I thought that Airbnb is recommending to assign static properties on a class after the class declaration ([Link](https://github.com/airbnb/javascript/tree/master/react#ordering)) instead of using the class static properties proposal?

mikefowler:
Sounds good, y'all, I'll move this back to using rest properties.

@charpeni: good call, this isn't in our styleguide. We've been using them a bit internally since React Native supports them. I'll let you and @lelandrichardson decide as collaborators on this project.

mikefowler:
- Removed lodash dependency in favor of rest properties.
- Pull propTypes out of static property

Feel free to review again, y'all.

aksonov:
@mikefowler Looks good, could you merge with latest master and submit again (now it has conflicts) ?

src/Router.js
```diff
@@ -6,63 +6,105 @@
...
-export default class extends Component {
+const {
+  RootContainer: NavigationRootContainer,
+} = NavigationExperimental;
+
+const propTypes = {
+  dispatch: PropTypes.func,
+};
+
+export default class Router extends Component {
```
charpeni: Since `propTypes` is assigned after the class declaration, `export default` should be moved to the end of the file.

```javascript
Router.propTypes = propTypes;

export default Router;
```

mikefowler:
@aksonov, conflicts resolve. @charpeni, addressed your comment as well. Should be good to merge.

charpeni:
LGTM, good work !

PR #605, opened 2016-04-27
Fix eslint errors in DefaultRenderer.js

mikefowler:


charpeni:
Maybe the render methods could be defined in the same order they are binded ?

```javascript
this.renderCard = this.renderCard.bind(this);
this.renderScene = this.renderScene.bind(this);
this.renderHeader = this.renderHeader.bind(this);
```

mikefowler:
Cool with me, @charpeni. Updated.

PR #608, opened 2016-04-27
Fix for navigation Actions not returning true or false.

bburnham:
This will allow for Actions.pop(), Actions.jump(), Actions.refresh(), Actions.focus() to return if the action was successful or not.


cridenour:
LGTM

bosung90:
Looks good to me too.

PR #609, opened 2016-04-27
Fix eslint errors in State.js

charpeni:


src/State.js
```diff
@@ -30,20 +30,27 @@ function getStateFromScenes(route, scenes, props) {
   return result;
 }
 
-export function getInitialState(route:{string: any}, scenes:{string:any}, position = 0, props = {}) {
+export function getInitialState(route: {string: any}, scenes:{string:any}, position = 0,
```
mikefowler: If this is hitting the line-length rule, let's just break the arguments to separate lines:

```js
export function getInitialState(
  route: {string: any}, 
  scenes:{string:any}, 
  position = 0, 
  props = {}
) {
  // ...
}

charpeni: You're totally right, I'll fix this tomorrow in the morning.

charpeni:
Pull request has been updated.

PR #610, opened 2016-04-27
Nicer readme.

r0b1n:
Next step is to reorder properties, because now they are messy. Combine they by types/purposes ( *button related, sceneStyle related, etc)


cridenour:
While you're in there, might as well make the 0.22 requirement more prominent (at least bold, maybe moved to top) to avoid a lot of confusion.

PR #618, opened 2016-04-29
Fix Reset and Replace

Zidail:
React-Native: 0.24.1
Compatible with 0.22+

Updates existing state children with new scenes so that it forces a re-render of the scene rather than removing the object which leaves the previous scene in the view.

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -95,12 +95,24 @@ function inject(state, action, props, scenes) {
       if (state.children[state.index].sceneKey === action.key) {
         return state;
       }
-      return { ...state,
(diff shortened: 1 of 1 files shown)


charpeni:
LGTM

JamesWatling:
Are we able to get this merged and approved? Really wanting to be able to upgrade to 0.24 but don't want to use my own private fork - and this solves my issues

charpeni:
I'll merge since no one else did review. Thank you for your work.

rturk:
This nice, however goes back to other discussions that we need to improve tests

PR #653, opened 2016-05-09
Send all props through to renderNavigationBar and renderTitle through props

mat2maa:
I have a custom extension to the default NavBar through `renderNavigationBar(props)` which, in turn, renders a custom title by passing `renderTitle` through a prop, leading to the change to NavBar.js.

In addition, there is the need for the `NavigationSceneRendererProps` to be passed to the custom title component to use the `AnimatedValue: position` within the `AnimatedView` (see: [here](https://gist.github.com/brentvatne/52af349a6b6ef2ee1b06#navigationheader)).


aksonov:
Looks good!

Issue #567, opened 2016-04-21
Code style linter and CI


charpeni:
Hi there,

I'm just wondering, should we add a linter in this repo ? For the sake of consistency between multiple contributors.

I suggest the [Airbnb React/JSX Style Guide](https://github.com/airbnb/javascript/tree/master/react) and it can be setup with [eslint-config-airbnb](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb), I can do that if you want to.

Furthermore, I think this repo should use a CI like [CircleCI](https://circleci.com/) it could automatically run eslinter and unit tests on each pull requests.

______

- [x] Implement eslint with [eslint-config-airbnb](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb) - (PR #574)
- [x] Fix eslint issues - #600 #601 #602 #603 #604 #605 #606 #609 #614 #625 
- [x] Add [mention-bot](https://github.com/facebook/mention-bot) to WebHooks
- [x] Add CI tests with eslint and mocha
- [x] Link with [CircleCI](https://circleci.com/)

aksonov:
Please do, thanks.

charpeni:
@aksonov  In addition, to improve the pull requests reviewing, i'm thinking about using the [mention-bot](https://github.com/facebook/mention-bot). Thoughts?

aksonov:
Looks interesting!

cridenour:
@charpeni You might want to watch out for doing both at once or every pull request is going to have you mentioned if you commit the new linted code :)

charpeni:
Oh, you're totally right.

The linking with [mention-bot](https://github.com/facebook/mention-bot) and [CircleCI](https://circleci.com/) will needs the interaction of a collaborator with access to the WebHooks.

cridenour:
True. I'd say, submit a PR with the new .eslintrc and such (which I agree would be an awesome addition), but let @aksonov do the big "reformat" commit.

charpeni:
I will do the fix to the eslint problems as mentioned in #574.

@aksonov Any reason why the tabulation is not consistent ?

aksonov:
No reason. Unfortunately webstorm sometimes doesn't do formatting

Pavel.

aksonov:
I've fixed some problems using --fix, but there is still many. Btw, strange that 'codacy' didn't found any, and requires usage of ", but aslant requires ' ;)

r0b1n:
[...]

PR #655, opened 2016-05-09
Fix eslint src/Navbar.js

charpeni:
Related to #567

### DIFF
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -122,6 +122,7 @@ const propTypes = {
   titleStyle: PropTypes.object,
   position: PropTypes.object,
   navigationBarStyle: PropTypes.object,
+  renderTitle: PropTypes.any,
 };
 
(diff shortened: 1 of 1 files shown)


aksonov:
@charpeni ops, sorry, enabled os x plan but it didn't run because it is paid! for open source! reverted back, could you submit another PR to run build again? Maybe we must use travisci instead with os x support for free (we could build Example project and run GUI tests for it!)

charpeni:
TravisCI is really slow, but I like your idea. Maybe we could setup TravisCI only for OS X build and keep CircleCI for Eslint and Mocha ?

aksonov:
Ugh, circleci schedules new builds, and don't want to remove 'os x' ones. I'm removing it and re-adding the project again...

aksonov:
Yes, we could try both

Issue #664, opened 2016-05-12
current master branch navigationState conflict on RN 0.24.1


jamesone:
As I was testing the `master` branch everything seemed to be working fine until I noticed one critical problem when changing scenes using `type: push`.

I pinpointed the issue to this snippet of code in the `DefaultRenderer.js` file:
```
// (Starting at line 62)
dispatchFocusAction({ navigationState }) {
    if (!navigationState || navigationState.component || navigationState.tabs) {
      return;
    }
    console.log ("\nNavigation state children before error: ", navigationState)
    const scene = navigationState.children[navigationState.index];
    Actions.focus({ scene });
  }
```

I noticed that my console.log was printing twice in a row before the following error occured:

`ExceptionsManager.js:61 navigationState.children[1].key "scene_position_feed" conflicts withanother child!`

For some reason the same `key` (position_feed) was set twice.

## Code:

I was pushing this scene:

```
 <Scene key="showPost" component={Post} title="Post"
                  navigationBarStyle={styles.navStyle}
                  titleStyle={styles.titleStyle}
                  sceneStyle={{top: 64}}
                  leftButtonTextStyle={styles.titleStyle}
                  rightButtonTextStyle={styles.titleStyle}
  />
```

From the feed scene:

```
 <Scene key="feed"
                    type="push" 
                    sceneStyle={{top: 64, flex: 1,}}
                    component={FeedLayout} title="Feed" rightTitle="Post"
                    onRight={()=> Actions.post()}
                    leftTitle="Map" onLeft={ () => Actions.map({all: true,})}
                    navigationBarStyle={styles.navStyle}
                    titleStyle={styles.titleStyle}
                    leftButtonTextStyle={styles.titleStyle}
                    rightButtonTextStyle={styles.titleStyle}
  />
```

If there's anything else I can add in here to help, let me know.

jamesone:
#636 Branched off from this issue

PR #665, opened 2016-05-12
Fix for #664 navigationState issue on RN 0.24.1

jamesone:
When pushing a new scene, there was an issue where the `scene.key` (`navigationState.children[navigationState.index].key`) would be conflict with the previous `scene.key` value.

What my piece of code does is make sure that the current `scene.key` is != the previous `scene.key`, if they're the same, then I add the `navigationState.index` to the end of the `scene.key` value, resulting in no conflict errors being thrown. 

Refer to this issue for more details:
#664

### DIFF
--- a/src/DefaultRenderer.js
+++ b/src/DefaultRenderer.js
@@ -64,6 +64,14 @@ export default class DefaultRenderer extends Component {
       return;
     }
     const scene = navigationState.children[navigationState.index];
+    // (temp or production?) Fix for #664
+    if (navigationState.index == 1) {
+      // Make sure the key names are not the same, or error will be thrown.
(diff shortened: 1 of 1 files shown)


rturk:
@jamesone This is interesting. Did you tested this with RN25 & RN26? As far as I know RN24 Navigation bugs are related to RN NavigationExperimental itself, not with RNRF.

jamesone:
@rturk  - I didn't test with RN25/26, only 0.24.1. The project I'm working on will need to be updated to work on RN V > 24. Is there a possibility you could test it?

If not I'll try convert my proj to work with a RN V > 24!

jamesone:
@joenoon has a better solution over at #667 - Turned out to be a typo.

PR #667, opened 2016-05-12
was literal 'position' a typo?

joenoon:
I'm not sure, so if someone could, please review if this is a safe change.

I think this is related to https://github.com/aksonov/react-native-router-flux/issues/664 and might be the right fix instead of https://github.com/aksonov/react-native-router-flux/pull/665 if the way I'm thinking about it is correct.

### DIFF
--- a/src/State.js
+++ b/src/State.js
@@ -42,7 +42,7 @@ export function getInitialState(
     return {
       ...scenes.rootProps,
       ...route,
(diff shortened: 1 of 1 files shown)


jamesone:
Can confirm that this fixed my issue. It is also a better fix than my #665 PR!

charpeni:
Oops, my bad. Sorry !

PR #682, opened 2016-05-16
Fix propTypes in NavBar

leoz:
Fix warnings related to type mismatch.

### DIFF
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -114,10 +114,10 @@ const styles = StyleSheet.create({
 
 const propTypes = {
   navigationState: PropTypes.object,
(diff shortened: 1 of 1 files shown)


sarovin:
Oops, i created a duplicate request.
@leoz can you change titleStyle PropTypes with this? 
`titleStyle: PropTypes.oneOfType([PropTypes.array, PropTypes.object])`
so we can use an array for style without warning.

sarovin:
#688

leoz:
Hi @sarovin , I guess our patches could be applied one on top of another as they fix different properties.

charpeni:
Introduced by #706.

PR #689, opened 2016-05-17
Fix bug hideNavBar not working

donnguyen:
The place we put hideNavBar checking just makes no sense so move it above the other rendering code lines

### DIFF
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -47,15 +47,17 @@ export default class NavBar extends React.Component {
         while (selected.hasOwnProperty("children")) {
             selected = selected.children[selected.index]
         }
+
+        if (child.hideNavBar || selected.hideNavBar) {
+            return null;
+        }
+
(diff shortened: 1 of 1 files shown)


cridenour:
The tests failed in an upstream library - weird. This LGTM.

charpeni:
@cridenour This is because the branch `0.24` doesn't contain a `circle.yml`. In fact, that commit should not rather be merged in master?

cridenour:
@charpeni Good catch. Should this be master or 0.26-wip?

joenoon:
This wont apply to 0.26-wip since the check is done quite a bit differently there

joenoon:
Closing since 3.24 is no longer supported.  Please reopen against master if there is still an issue.

eranbes:
hideNavBar={true} does not work in 0.26. 
I have this
 <Scene key="startup" initial={true} hideNavBar={true} /> yet it shows the navbar

PR #696, opened 2016-05-18
Force assert dependencies to version 1.3.0

Glavin001:
Using assert version 1.4.0 generates "Unable to resolve module buffer..." error.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     }
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


charpeni:
See #697 #698.

Glavin001:
Excellent. #698 is the way to go. Thanks for your quick response and merging!

Issue #697, opened 2016-05-18
Unable to resolve module buffer


joenoon:
@ronak301 ran into this on gitter and I tracked it down to a recent release of "assert".  I've created an issue over there: https://github.com/defunctzombie/commonjs-assert/issues/20

For now, I've updated the 0.26-wip branch to lock "assert" @ 1.3.0 : https://github.com/aksonov/react-native-router-flux/commit/15696515400126b74ce8ff93ca42ec558be58d44

Any ideas how to address it in the existing versions?

The hotfix is for the user to manually force their own dependency:

```
npm install --save assert@1.3.0
```

So maybe that needs to go in the README on master?  Its unclear if this is just a temp bug upstream and if they plan on yanking 1.4.0 or if we need to plan around this.

ronak301:
Ya this have solved my problem for the time being. Thanks to @joenoon for quick reply.

JimmyStavros:
I'm still having the same issue even after manually installing the dependency, anything else I can try?

joenoon:
@JimmyStavros try a thorough reset:

```
(set -ex; rm -rf node_modules/ && npm cache clear && npm install && (watchman watch-del-all || true) && rm -fr $TMPDIR/react-* && node_modules/react-native/packager/packager.sh --reset-cache true)
```

aksonov:
@joenoon what about removing this dependency at all and replace to own function to check condition and raise exception if it is not true?

joenoon:
@aksonov that sounds great!

calvinmetcalf:
FYI this exists https://github.com/calvinmetcalf/minimalistic-assert

aksonov:
@calvinmetcalf Let's don't repeat the same error - avoid dependency if possible. Because author of minimalist-assert could publish new version that brake build again... It is definitely weak side of NPM.

calvinmetcalf:
I'm the author of minimalistic-assert fyi but it would probably make sense to just require a specific version

cridenour:
I think @joenoon pushed a version to our new Utils file. But let's all remember the left-pad insanity before we introduce dependencies for 3 lines :)

lwansbrough:
Libraries should version lock their dependencies. Using the carat is dangerous.

calvinmetcalf:
in fairness it's especially dangerous for a situation like react-native
where you are consuming libraries from a different ecosystem meaning things
that are not a breaking change in node/browserify/webpack are breaking for
you guys

PR #698, opened 2016-05-19
remove dependency on assert

joenoon:
@aksonov let me know if you are good with this, per https://github.com/aksonov/react-native-router-flux/issues/697


anthonywebb:
This would solve the issue I am bumping up against (#697) hoping it is merged

PR #720, opened 2016-05-23
Enable the popping of multiple scenes

rseemann:
Extend the pop capability to pop multiple scenes at once, this can be used to pop a modal screen with more than scene pushed into the modal's navigation stack.

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -46,11 +46,12 @@ function inject(state, action, props, scenes) {
             case POP_ACTION2:
             case POP_ACTION:
                 assert(!state.tabs, "pop() operation cannot be run on tab bar (tabs=true)")
+                const scenesToPop = action.data || 1;
                 return {
                     ...state,
-                    index:state.index-1,
-                    from: state.children[state.children.length - 1],
-                    children:state.children.slice(0, -1),
+                    index:state.index - scenesToPop,
+                    from: state.children[state.children.length - scenesToPop],
+                    children:state.children.slice(0, - scenesToPop),
                 };
             case REFRESH_ACTION:
                 return {


joenoon:
@rseemann I think the feature request makes sense - to be able to pop back to a certain point in the navigationState.  But this looks a little arbitrary - just accepting a number to cut from the end and decrement the index, seems like there would be a pretty good chance of corrupting the navigationState by accident.

I was thinking this would be more along the lines of: "pop back to scene foobar" and the navigationState would be walked backwards until the foobar scene is found, and then everything in front of it would be dropped.

I'm not sure though.  @aksonov any thoughts?

rseemann:
That was actually the main goal, but the first step to do it, imho, would
be to allow popping of more than one scene at a time.

How could this corrupt the navigationState and how could it be done
differently other and changing the values the state as they are being
changed now?

joenoon:
Currently here: https://github.com/aksonov/react-native-router-flux/blob/2d84f885a6723d3a0dfba18a4db6554e6260581f/src/Reducer.js#L61

Its impossible to end up with an invalid index, because there is a check to protect against decrementing beyond 0, and otherwise the decrement is 1.

If the user is allowed to send whatever number they want, then they could decrement the index by 4 on accident/user-error, and end up with an index of -3.  So at the very least there needs to be an assert protecting against that scenario.

rturk:
@rseemann I agree with @joenoon this is an nice feature, however we need a robust code. Any app that uses Sagas or handle push notifications can easily have the index corrupted.

aksonov:
Closing it for now, @rseemann feel free to submit fixed PR and ask @joenoon to review.

PR #721, opened 2016-05-23
propagate selected

alfonsodev:
propagates the selected property to the child element, 
allows to check if the tab is selected within the component itself.


alfonsodev:
Closing as is not working as intended, help is welcome .

Issue #717, opened 2016-05-22
Tapping a tab should navigate to the default scene for that scene group


Olliepop:
### Version
- react-native-router-flux v3.26.1
- react-native v0.26.0 & v0.26.1

I am using the Example app.
### Expected behaviour

In the Example, when I go to the "TabBar page" then "push new scene" then tap the tab i'm currently in ("Tab #2" for example) i expect my current scene to pop so that i'm back at the default scene for that tab.

Because the action type `jump` with key `tab2` is dispatched, i expect to be taken to the default scene for that tab (tab2_1)

This is the group I am talking about:

`<Scene key="tab2" initial={true} title="Tab #2" icon={TabIcon}>
    <Scene key="tab2_1" component={TabView} title="Tab #2_1" renderRightButton={()=><Right/>} />
    <Scene key="tab2_2" component={TabView} title="Tab #2_2" onLeft={()=>alert("Left button!")} leftTitle="Left" duration={1} panHandlers={null}/>
</Scene>`
### Actual behaviour

3 actions are dispatched: 
- Jump (key: tab2)
- Push (scene.key: 0_modal)
- Push (scene.key: 1_tabbar)

Despite these actions, no actual navigation is performed and the user remains on the same child scene of a tab that they were on. 
### Steps to reproduce
1. Open the RNRF `Example` app
2. Tap "Go to TabBar page"
3. Tap "push new scene"
4. Tap "Tab # 2" in the tab bar

I expect that # 4 takes us back to scene `tab2_1` instead of staying on the current scene.

cridenour:
I think this is a good idea, would love to see a PR with it if you're up to it!

Olliepop:
@cridenour Cool - i thought this was a bug with intended functionality because of the actions dispatched when you click a tab. But if it's an enhancement i'd definitely be happy to work on it.

cridenour:
I believe those actions are only dispatched because we send them regardless of whether the tab is selected.

joenoon:
Agree this would be nice - just checked the Facebook and Twitter apps and they behave as described above

Olliepop:
Question related to solving this issue:

I had this solved in 3.22.23 by chaining `Actions.pop()` to the number of children in the tab scene navigationState. 

In the latest implementation of TabBar within the `onSelect` method we're unable to execute multiple Actions i.e:
```
Actions.scene_1();
Actions.scene_2();
Actions.scene_3();
```

[...]

PR #729, opened 2016-05-24
UPDATE: Pop tab children when tab is selected

Olliepop:
Fixes https://github.com/aksonov/react-native-router-flux/issues/717 by adding a prop (default: false, see updated docs) `popChildrenOnSelect`

If the prop `popChildrenOnSelect` is set to true on the tab child Scene, then pop all children in the stack to go back to the initial scene for this tab when it is selected again.

I have also fixed an issue where React `15.1.0` was being updated in new installs of Example, causing an error `super expression must either be null or a function not undefined`

### DIFF
--- a/Example/package.json
+++ b/Example/package.json
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "react": "15.0.2",
-    "react-native": "^0.26.1",
+    "react-native": "^0.26.0",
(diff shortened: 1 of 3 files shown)


Olliepop:
`Actions.pop()` isn't working in Example so closing until i fix that

PR #737, opened 2016-05-26
Fix Typo in Switch referring to this.props

SirensOfTitan:
It looks like this reference was missed in the conversion to a functional component.

### DIFF
--- a/src/Switch.js
+++ b/src/Switch.js
@@ -13,7 +13,7 @@ export default function Switch(props) {
   if (!navigationState) console.error(`Cannot find a scene with key “${selectedKey}”`);
   return (
     <DefaultRenderer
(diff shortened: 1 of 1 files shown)


aksonov:
Thanks!

charpeni:
@SirensOfTitan It breaks the CI, can you fix this?

aksonov:
@charpeni Looks like i've broken it during my previous commit. What kind of command i should run locally to prevent that? I've run npm run eslint, but it outputs a lot of errors always.

charpeni:
@aksonov You're right, I added Example folder into eslint scripts but not into CI since it contains a lot of errors. I'll do a pull request to remove Example folder from `npm run eslint`.

PR #739, opened 2016-05-26
Add prop to override the tabBar shadow style

laurentnguyen42:


charpeni:
@laurentnguyen42 It breaks the CI, can you fix this?

aksonov:
@charpeni Looks like i've broken it during my previous commit. What kind of command i should run locally to prevent that? I've run npm run eslint, but it outputs a lot of errors always.

joenoon:
The previous tab implementation is coming back soon which will wipe this out, but merge this for now

PR #742, opened 2016-05-27
Update README.md - Include RN Compatibility table

rturk:
We continue to receive new issues with questions/issues related incompatible versions. Folks open issues without reading legacy issues.. 

Looks like this will continue for some time given that React Navigation core is not mature and we can expect more breaking changes in the upcoming RN releases

Hopefully including this table in the main Read.me will reduce number of new issues


cridenour:
I don't believe 3.23 and 3.24 were ever "supported" as we closed all issues with them. I would put them in the same category as 3.25.

rturk:
@cridenour makes sense. Actually there are only a few cases that worked, most with tons of hacks

PR #743, opened 2016-05-27
WIP old tab bar

joenoon:
This PR brings back the previous tab implementation and expands on it to work with RN 0.26 (gets rid of react-native-tab-navigator)

- Also adds some render performance improvements using StaticContainer
- Avoids rendering tabbed scenes until they are selected for the first time

- Locks versions more tightly

BROKEN:

- I'm having trouble figuring out the best way to let the user style their scenes.  There needs to be a way to know if the nav and tab bar is hidden or present, to allow for adding margin, etc.  The best place to observe the problem is the Example app: Tab Bar > push new scene > any of these options.  The red border should be visible for any combination.  If anyone has ideas or wants to add to the PR that would be very helpful.

### DIFF
--- a/Example/Example.js
+++ b/Example/Example.js
@@ -48,11 +48,11 @@ const reducerCreate = params=>{
 };
 
 // define this based on the styles/dimensions you use
-const getSceneStyle = function (props) {
+const getSceneStyle = function (/* NavigationSceneRendererProps */ props) {
   return {
     flex: 1,
-    marginTop: props.hideNavBar ? 0 : 64,
-    marginBottom: props.hideTabBar ? 0 : 49.5,
+    // marginTop: props.scene.navigationState.hideNavBar ? 0 : 64,
+    // marginBottom: props.scene.navigationState.hideTabBar ? 0 : 50,
     backgroundColor: '#fff',
     shadowColor: null,
     shadowOffset: null,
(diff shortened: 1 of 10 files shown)


joenoon:
Also broken:  seems the inner navbars are now rendering underneath the outer navbars.  So back button etc isnt showing up... will try to figure out why today.  And if anyone else spots the issue please let me know

joenoon:
I think I have this working as expected.

It would be great if I could get some extra eyes on this.  There were lots of issues along the way with getting the scene styling and hideNavBar/hideTabBar values passed through to getSceneStyle.  And I also ran into multiple navbars that I didn't know were previously there by accident when playing with margins.  So I'm not quite sure if some of the things worked at some point before just by luck, or if I've introduced a bug or misunderstood how some of the navigation should function.

The result was pretty much trial and error in the example app to figure out the right combination that produced the right result.  

If we could get a few people confirming the Example app works as they expect, AND they can confirm their real app works as expected, that would be awesome!

rturk:
@joenoon I'll try to test this week

Cordazar:
Tested it out in a real project. The tabBar is working and other issues with double navbars and stuff are also solved.

joenoon:
Just to keep track of what would probably need to go into release notes for this, and for anyone testing this:

- check updated `getSceneStyle` signature in Example app
- set hideNavBar hideTabBar on the root scene

aksonov:
It works well for me as well, thanks! I'm merging

aksonov:
@joenoon omg, just noted that all my navbar buttons disappeared :( I'm using leftButton, rightButton properties.

PR #758, opened 2016-05-30
Fix tabBarStyle for hideTabBar

q6112345:
For the hidden TabBar still clickable problem.


aksonov:
@q6112345 Is it still issue with latest version?

charpeni:
@q6112345 If it still an issue with the latest version, please rebase and fix conflicts.

PR #774, opened 2016-06-03
add typings

bang88:
Hello

Can i add a typing for RNRF,Thanks.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,303 @@
+// typings for react-native-router-flux@3.26.16
+// created by bang88 (https://github.com/bang88)
+
+import * as React from 'react';
+
+declare namespace RNRF {
+
+  export interface SceneProps {
+    // Scene
+    /**
+     * Will be used to call screen transition, for example, Actions.name(params). Must be unique.
+     */
+    key: string,
+    /**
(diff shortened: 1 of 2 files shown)


rturk:
Hi @bang88 are you also plaining to maintain this? My sense is that this will break quite often given the number of changes that React and RNRF will experience in the upcoming releases.

charpeni:
@rturk Totally agree.

I don't know for the other collaborators, but I'll prefer [Flow](http://flowtype.org/) type checker instead of TypeScript.

bang88:
@rturk  That's fine,but i think if you guys found an issue in typings you can improve it,not just me.
@charpeni  I don't want choose Flow or typescript,just give this a try for lntellisense when use https://code.visualstudio.com/

Accept or not your choice. I am free.

joenoon:
I like this even though it means a little more maintenance.  It can only really be maintained by people who use vscode or something else that interprets these typing files.  So it might break or be outdated from time to time, but thats better than not including it IMO.

rturk:
@joenoon makes sense. Worse case scenario this will only affect folks using vscode, and it will be always another tool to validade the code.

PR #791, opened 2016-06-06
790: Avoid rightButtonImage or rightTitle warnings when set to null

vpasquier:


vpasquier:
### Version
Tell us which versions you are using (react native 0.26 is required for v3): 

- react-native-router-flux v3.22.0
- react-native v0.22.2

### Expected behaviour

Not having warnings when setting `null` to `rightButtonImage` or `rightTitle`

### Actual behaviour

The opposite

### Steps to reproduce

Explained in the expected behaviour

Another solution would be hide and display those buttons, based on rules.

charpeni:
LGTM. I'll let someone else do the merge to be sure that wasn't intentional.

vpasquier:
Merci :)

PR #793, opened 2016-06-07
[actions]: add support when dealing Scenes defined using array map

alextkd:
In the render function when defining the Scenes, if you use 
{ myViewsArray.map((item, index)=>this.buildNavScenes(item, index)) } it work with this small fix, without needing to enclose this into an additional Scene


charpeni:
Please fix the eslint errors.

You can test it locally with `npm run eslint`.

alextkd:
It is weird because i haven't changed any of that. I'm looking on fix that now.

Issue #39, opened 2015-11-18
Go back twice?


nbastoWM:
I want from a button to navigate back twice.
I tried Actions.pop() twice but with no success.

aksonov:
Try Actions.pop(2)

wootwoot1234:
@aksonov  This should be added to the documentation, or did I miss it?

PR #798, opened 2016-06-08
[Feature/Action] Enable the popping of multiple scenes

rseemann:
Create the `popToParent()` Action that takes a parent name as argument and pops the scenes until the targeted parent is the current scene in place.

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -12,6 +12,7 @@
 import {
   PUSH_ACTION,
   POP_ACTION2,
+  POP_TO_PARENT_ACTION,
   JUMP_ACTION,
   REPLACE_ACTION,
   RESET_ACTION,
@@ -56,18 +57,22 @@ function inject(state, action, props, scenes) {
     return state;
   }
   let ind;
+  let numPops = 1;
   switch (action.type) {
+    case POP_TO_PARENT_ACTION:
+      numPops = action.numPops;
     case POP_ACTION2:
     case POP_ACTION:
       assert(!state.tabs, 'pop() operation cannot be run on tab bar (tabs=true)');
       if (state.index === 0) {
         return state;
       }
+
       return {
         ...state,
-        index: state.index - 1,
+        index: state.index - numPops,
         from: state.children[state.children.length - 1],
-        children: state.children.slice(0, -1),
+        children: state.children.slice(0, - numPops),
       };
     case REFRESH_ACTION:
       return props.base ?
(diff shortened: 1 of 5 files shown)


rseemann:
@joenoon what do you think of this way of implementing the multiple pops?

joenoon:
@rseemann cool!  This looks like it makes sense.  The one thing I'm unsure of is the state.children search - what if the child hierarchy is more complicated?  Does this need to walk the full tree instead of only direct children?  I'm not sure.  Would be awesome if you could add something to the Example app to demo the intended scenario - that would open up a place to experiment with more popToParent scenarios.

rseemann:
@joenoon I also thought about a more complex structure, but I wouldn't know how to handle it, since I don't fully understand the routing structure. Maybe I'll do some testing with the Example app as you mentioned, great idea.

aksonov:
@rseemann Route structure is very simple - it is just tree, with children as scenes. Child could have other children, etc.

aksonov:
@rseemann could you also fix eslint errors?

rseemann:
@aksonov The structure I kind of understand, popping those, instead, I'm not sure I understand, specially  with the index number. About the linting, I'll fix those, for sure.

aksonov:
@rseemann index number is currently selected scene.

rseemann:
@aksonov so this number does not depend on how deep the structure is, but rather on how many scenes there are currently in the stack?

aksonov:
@rseemann yes. the same as in NavigationExperimental structure.

aksonov:
For 'tabs', it represents number of selected tab.

rseemann:
@aksonov that seems simple enough then, I'll make the necessary changes, add some examples and fix the linting then. Thanks for the explanations.

src/Reducer.js
```diff
@@ -56,18 +57,22 @@ function inject(state, action, props, scenes) {
     return state;
   }
   let ind;
+  let numPops = 1;
```
Swordsman-Inaction: Sorry, I didn't see yours before I create the PR #806 , I write something similar to your `numPops` . According this issue #39 , I think old version `Actions.pop()` could do `Actions.pop(2)` things, so I implemented this with several `assert` checking.

rseemann:
@joenoon and @aksonov I made some much deeper changes to handle the navigation to any scene that exists in the stack, node or component. Both usages can be seen at the example, with the newly component login3. I'm not fully happy with the node/component detection, maybe you can tell me if that's a reliable check.

BenBenz:
@rseemann  popToParent() works at this time ?

rseemann:
@BenBenz I haven't checked on this since a while. Unless something changed in the navigation way, it should work.

BenBenz:
@rseemann Thanks for this information 👍

PR #801, opened 2016-06-09
[tabs]: add custom selected item tab style

alextkd:
add support to customize a selected tab item in case you may want to have custom background for selected tab items

### DIFF
--- a/src/Scene.js
+++ b/src/Scene.js
@@ -15,6 +15,7 @@ export default class extends React.Component {
 
   static propTypes = {
     tabBarStyle: View.propTypes.style,
+    tabBarSelectedItemStyle: View.propTypes.style,
     tabBarShadowStyle: View.propTypes.style,
     tabSceneStyle: View.propTypes.style,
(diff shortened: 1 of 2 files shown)


Crash--:
Had an issue after upgrading to the 3.26.30 because of this commit. I've a customized `tabBarStyle` and I want this style applied if I haven't specified a tabBarSelectedItemStyle for the selected item not the 'white' background. Can we replace the current line by this one or do I need to add tabBarSelectedItemStyle to every tabbar I use? 

```jsx
selectedIconStyle={[{ backgroundColor: 'white' }, state.tabBarStyle, state.tabBarSelectedItemStyle]}
```

joenoon:
I had the same experience and I had to null out the backgroundColor.  I don't think we need the default styling at all.  What about just:
```
            style={state.tabBarStyle}
            selectedIconStyle={state.tabBarSelectedItemStyle}
```

joenoon:
PR here: https://github.com/aksonov/react-native-router-flux/pull/839

PR #806, opened 2016-06-10
Feature(multiple pop): now we can pop multipe scenes with Actions.pop

Swordsman-Inaction:
now we can pop multipe scenes with Actions.pop([number]), like Actions.pop(2)

But I don't know much about the mechanism, please tell me if I miss something.

@aksonov

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -70,7 +70,7 @@ function inject(state, action, props, scenes) {
     }
 
     case POP_ACTION2:
-    case POP_ACTION:
+    case POP_ACTION:{
       assert(!state.tabs, 'pop() operation cannot be run on tab bar (tabs=true)');
       if (state.index === 0) {
         return state;
@@ -86,7 +86,7 @@ function inject(state, action, props, scenes) {
         assert(popNum > 1,
           'The data is the number of scenes you want to pop, it must be bigger than 1.');
         assert(popNum <= state.index,
-          "The data is the number of scenes you want to pop, " + 
+          'The data is the number of scenes you want to pop, ' +
           "it must be smaller than scenes stack's length.");
       }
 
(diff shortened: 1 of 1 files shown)


src/Reducer.js
```diff
@@ -63,11 +63,21 @@ function inject(state, action, props, scenes) {
...
         return state;
       }
+
+      let popNum = 1
+      if (action.data) {
+        assert(typeof(action.data) === 'number', 'The data is the number of scenes you want to pop, it must be Number')
+        popNum = action.data
+        assert(popNum % 1 === 0, 'The data is the number of scenes you want to pop, it must be integer.')
+        assert(popNum > 1, 'The data is the number of scenes you want to pop, it must be bigger than 1.')
+        assert(popNum < state.children.length, "The data is the number of scenes you want to pop, it must be smaller than scenes stack's length.")
```
aksonov: Probably you need to check against state.index, not state.children.length...

Swordsman-Inaction: @aksonov  changed

Swordsman-Inaction:
@aksonov In the `Reducer.js`, before `update` is called, `action.parent` will be assigned, But neither `update()` or `inject` use it. So does it matter? If it matters, I think I should update the implement, since the `action.parent` is wrong if you are popping several scenes.

aksonov:
@Swordsman-Inaction it doesn't matter actually, if your PR changes state properly without settings action.parent. Could you fix conflicts and eslint errors?

joenoon:
I think this can be closed since https://github.com/aksonov/react-native-router-flux/pull/798 was merged?

Swordsman-Inaction:
@joenoon @aksonov  #798 is a good feature. But this one is trying to get this (https://github.com/aksonov/react-native-router-flux/issues/39#issuecomment-157990785) back.
Both of them are basically trying to do the same thing. If you guys think #798 is enough, then this one can be closed

joenoon:
@Swordsman-Inaction ok then I can see how this makes sense.  Besides the eslint and conflicts, the only other thing that comes to mind is using `action.data`.  Should that be so generic, or should it be `action.popNum` or something?

Swordsman-Inaction:
@joenoon `action.popNum` looks great. I'll fix those

Swordsman-Inaction:
@aksonov @joenoon Done

PR #818, opened 2016-06-14
Alignment the elements for the navbar

sarovin:
I think that the elements are not well aligned now.
This fix it for me. Do you agree?

### DIFF
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -52,7 +52,7 @@ const styles = StyleSheet.create({
         top: 20,
       },
       android: {
-        top: 15,
+        top: 5,
       },
     }),
     left: 0,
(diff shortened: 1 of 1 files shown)


src/NavBar.js
```diff
@@ -47,15 +47,29 @@ const styles = StyleSheet.create({
...
-    height: Platform.OS === 'ios' || Platform.Version > 19 ? 64 : 44,
+    ...Platform.select({
+      ios: {
+        height: 64,
+      },
+      android: {
+        height: 64,
+      },
+    }),
     right: 0,
```
aksonov: Are you sure that height should be the same for android/ios? We had issue with that...

sarovin:
There is a problem with the Translucent StatusBar.
We can't see if the translucent is set in StatusBar, no?

schermata 2016-06-15 alle 15 55 54

aksonov:
Okey, will you deal with possible issues after your PR ? ;)

sarovin:
Yes, no problem.

PR #820, opened 2016-06-15
Feature Hand Back Key for Android

Swordsman-Inaction:
For know, RNRF doesn't handle back kay on Android. App will exit anyway when you press the back key, no matter which scene you are at. This pull request handles the back key action. It will pop current scene when you press back key, and throw a error when you are on the root scene on Android, then the BackAndroid will cache the error and exit the app.

### DIFF
--- a/src/Router.js
+++ b/src/Router.js
@@ -10,6 +10,7 @@ import React, {
   Component,
   PropTypes,
 } from 'react';
+import { BackAndroid } from 'react-native'
 import NavigationExperimental from 'react-native-experimental-navigation';
 
 import Actions from './Actions';
@@ -97,6 +98,16 @@ class Router extends Component {
 
   render() {
     if (!this.state.reducer) return null;
+
+    BackAndroid.addEventListener('hardwareBackPress', () => {
+      try {
+        Actions.pop();
+        return true;
+      }
+      catch (err) {
+        return false;
+      }
+    })
 
     return (
       <NavigationRootContainer
--- a/src/Router.js
+++ b/src/Router.js
@@ -10,7 +10,7 @@ import React, {
   Component,
   PropTypes,
 } from 'react';
-import { BackAndroid } from 'react-native'
+import { BackAndroid } from 'react-native';
 import NavigationExperimental from 'react-native-experimental-navigation';
 
 import Actions from './Actions';
@@ -107,7 +107,7 @@ class Router extends Component {
       catch (err) {
         return false;
       }
-    })
+    });
 
     return (
       <NavigationRootContainer
--- a/src/Router.js
+++ b/src/Router.js
@@ -103,8 +103,7 @@ class Router extends Component {
       try {
         Actions.pop();
         return true;
-      }
-      catch (err) {
+      } catch (err) {
         return false;
       }
     });
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -75,9 +75,9 @@ function inject(state, action, props, scenes) {
       assert(!state.tabs, 'pop() operation cannot be run on tab bar (tabs=true)');
 
       if (Platform.OS === 'android') {
-        assert(state.index > 0, 'You are already in the root scene.')
+        assert(state.index > 0, 'You are already in the root scene.');
       }
-      
+
       if (state.index === 0) {
         return state;
       }
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -22,6 +22,7 @@ import {
 
 import { assert } from './Util';
 import { getInitialState } from './State';
+import { Platform } from 'react-native';
 
 // WARN: it is not working correct. rewrite it.
 function checkPropertiesEqual(action, lastAction) {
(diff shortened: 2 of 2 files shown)


rseemann:
The default behaviour for an Android app when the back button is clicked in the root view, is to exit the app, as far as I remember.

Swordsman-Inaction:
@rseemann  Really? I'm using the 3.26.16 version right now. But I'll check the latest version later. I close this PR firstly.

rseemann:
@Swordsman-Inaction I mean for native Android development, I'm not sure how it works here

Swordsman-Inaction:
@rseemann  I updated the PR's description.

esutton:
I am not sure you **always** want to exit without giving the developer a choice.

On Android, when hardware back button is pressed in the root view, it **normally** ( **but not always** ) exits the app.

On an app that user can drill-down many levels, you may want to display a toast prompt  in case user accidentally pressed back too many times. A typical toast prompt is "**Tap back once more to exit**".

The app File Manager HD(File transfer) does this.
   https://play.google.com/store/apps/details?id=com.rhmsoft.fm.hd&hl=en&gl=us

I handle Android in my main view like this.

```
import {
  Alert,
  BackAndroid,
  ListView,
  Platform,
  StyleSheet,
  ToastAndroid,
  Text,
  View
} from 'react-native';


  backAndroidHandler () {
    try {
      Actions.pop();
      return true;
    }
    catch (err) {
      // let backCount = this.state.backCount + 1;
      // let exitToHomeScreen = (1 < backCount) ? true : false;
      // if( false === exitToHomeScreen) {
      //   ToastAndroid.show("Tap back once more to exit." + backCount, ToastAndroid.SHORT);
      // } else {
      //   backCount = 0;
      //   BackAndroid.exitApp();
      // }
      // this.setState({
      //   backCount: backCount,
      // });
      // return !exitToHomeScreen;

      // Cannot get toast to disappear - Exit to home without Toast prompt
      return false;
    }
  },
```

I think RN has a problem with toast.  I could never get it disappear.  I gave up on the toast prompt.

Swordsman-Inaction:
@esutton Good point. But RNRF doesn't handle the back key at all right now. It just exits the app whenever you press the button. My way is not very good, should we discuss this and make a better solution? @aksonov

esutton:
In my non-expert-Android opinion, RNRF should not exit the app.

Let the developer do what he wants.  My backAndroidHandler() makes RNRF works the way I want it to.  

When user backs out to the root, the backAndroidHandler() in my MainView where the RNRF router lives get called and I can choose to exit or not.

timzaak:
```javascript
import { BackAndroid } from 'react-native';
import { Actions, Reducer } from 'react-native-router-flux';
import * as Toast from '../util/toast';
let hasInit = false;
let currentSceneName = null;

function getCurrentName(obj) {
  const { index, children, name } = obj;
  if (index !== undefined) {
    if (children) {
      return getCurrentName(children[index]);
    }
  }
  return name;
}
export const routerReducer = params => {
  return (state, action) => {
    const nextState = defaultReducer(state, action);
    currentSceneName = getCurrentName(nextState);
    return nextState;
  };
};
export function platformInit() {
  if (hasInit) {
    return;
  }
  hasInit = true;
  BackAndroid.addEventListener('hardwareBackPress', () => {
    // the root view
    if (['Login',
        'tab1',
        'tab2',
        'tab3',
        'tab4',
        'doctorTab1',
        'doctorTab2',
        'doctorTab3'].indexOf(currentSceneName) !== -1) {
      if (this.lastBackPressed && this.lastBackPressed + 3000 >= Date.now()) {
        BackAndroid.exitApp();
        return false;
      }
      this.lastBackPressed = Date.now();
      Toast.show('press again, exit');
      return true;
    }
    Actions.pop();
    return true;
  });
}

```


this is my code to handle the android back key.

lideen:
I think it is a bad idea to force the user to have this back button behavior. Maybe it is better to make this optional? In case you want a custom behavior as @esutton and @timzaak presented. I can also think of other use cases where you might want to show a warning or similar if the user is about to navigate back from a scene.

Swordsman-Inaction:
@aksonov @joenoon I think we should revert this PR, maybe we can get a better solution later.

joenoon:
It looks like something that would do well as a README and Example addition so people can hook it up if they want.  But, the event listener should not be added in the render, and would need to be cleaned up properly.

esutton:
I **really** like the built in pop that occurs on Android-back.

I think it is the automatic exitApp at the root, on Android-back  that developers may object to.

src/Router.js
```diff
@@ -98,6 +99,15 @@ class Router extends Component {
   render() {
     if (!this.state.reducer) return null;
 
+    BackAndroid.addEventListener('hardwareBackPress', () => {
+      try {
+        Actions.pop();
+        return true;
+      } catch (err) {
+        return false;
```
xseignard: Why not something like this instead this `return false;`:

```
return typeof this.handleAndroidExit === 'function' ? this.handleAndroidExit() : false;
```

Where the `handleAndroidExit()` is a function passed as a `props` to the `Router`. Like this people would be able to define it's custom exit behavior and/or rely to the default one.

This would do the trick I guess.

yanqiw:
@timzaak 's solution help me.

PR #823, opened 2016-06-15
Fix in Typescript definitions

dalcib:
Some fix in Typescript definitions:
- prop 'initial' in 'Scene' should be optional
- all props of scenes that could be used in 'Router' should be optional there
- optional wrapBy in 'Scene'
- optional props in Action.pop


bang88:
👍

PR #827, opened 2016-06-16
Tabbar Item onPress

phproot:
Use onPress open modal

Examples:
```
<Scene key="Modal" component={Modal}>
        <Scene tabs={true}  …  >
            <Scene key="Launch"  onPress={()=>Actions._Launch()} …/>
            <Scene key="Contacts" … />
            <Scene key="Me"  … />
        </Scene>
    <Scene key="_Launch" component={Launch}/>
</Scene>
```

### DIFF
--- a/src/DefaultRenderer.js
+++ b/src/DefaultRenderer.js
@@ -212,6 +212,15 @@ export default class DefaultRenderer extends Component {
     }
 
     if (SceneComponent) {
+      if (navigationState.tabs || ! navigationState.parent) {
+        return (
+          <SceneComponent
+            {...navigationState}
+            onNavigate={onNavigate}
+            navigationState={navigationState}
+          />
+        );
+      }
       return (
         <View
           style={[styles.sceneStyle, navigationState.sceneStyle]}
--- a/src/DefaultRenderer.js
+++ b/src/DefaultRenderer.js
@@ -12,7 +12,6 @@ import React, {
 } from 'react';
 import {
   Animated,
-  NavigationExperimental,
   View,
   StyleSheet,
 } from 'react-native';
@@ -21,7 +20,7 @@ import TabBar from './TabBar';
 import NavBar from './NavBar';
 import Actions from './Actions';
 import { deepestExplicitValueForKey } from './Util';
-
+import NavigationExperimental from 'react-native-experimental-navigation';
 const {
   AnimatedView: NavigationAnimatedView,
   Card: NavigationCard,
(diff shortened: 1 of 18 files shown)


joenoon:
Looks ok to me but I don't fully understand - maybe I haven't needed this myself yet.  Would it be possible to make the tab have its own `<TouchableOpacity>` that intercepted the tap if there is a need to do something (non-standard?) like this?  Is this just a more convenient alternative or are there other benefits?

Can you resolve conflicts and eslint?

As a side note (not about this PR), that `condition` and how this whole section of the code works gives me a headache.

src/Reducer.js
```diff
@@ -56,19 +58,51 @@ function inject(state, action, props, scenes) {
...
   }
   let ind;
+
   switch (action.type) {
+    case POP_TO: {
+      const targetIndex = action.targetIndex;
+
+      return {
+        ...state,
+        index: targetIndex,
```
alextkd: You may want to add from index as well

src/TabBar.js
```diff
@@ -25,7 +25,11 @@ class TabBar extends Component {
         `No action is defined for name=${el.props.name} ` +
         `actions: ${JSON.stringify(Object.keys(Actions))}`);
     }
-    Actions[el.props.name]();
+    if (el.props.onPress){
```
alextkd: can you make sure it is a function sent ( typeof el.props.onPress === "function")

src/DefaultRenderer.js
```diff
@@ -212,6 +212,15 @@ export default class DefaultRenderer extends Component {
     }
 
     if (SceneComponent) {
+      if (navigationState.tabs || ! navigationState.parent) {
```
alextkd: I'm wondering why you would need this. navigationState.tabs will never have a component attached to it and this condition is for navigationState.component, or maybe i got it wrong. 
<Scene key="my_tabs" tabs>
 .... your other scenes with components that represent tab views ... 
</Scene>

phproot: Has been canceled

src/TabBar.js
phproot: Well！

src/Reducer.js
phproot: Here I do not quite understand

src/TabBar.js
joenoon: Seems like a job for type checking with flow or typescript - not sure about going down the path of doing runtime type checking everywhere?  But cheers to anyone who would attempt passing a non-function to `onPress` :)

src/Reducer.js
joenoon: @alextkd could you expand on your comment here? I didn't understand either

src/TabBar.js
alextkd: People coming other frameworks where they could just put there the action on the class itself mentioned as a string, onPress='onMyButtonPress' where onMyButtonPress is a method defined within the class :) well i'm sure it will rarely happen, it is more as a sanity check for avoiding type missmatch exception. I do not have anything in mind merging this in like it is though.

PR #828, opened 2016-06-16
tabs show onPress item

phproot:
tabs show onPress item

Examples:

```
<Scene key="Modal" component={Modal}>
        <Scene tabs={true}  …  >
            <Scene key="Launch"  onPress={()=>Actions._Launch()} …/>
            <Scene key="Contacts" … />
            <Scene key="Me"  … />
        </Scene>
    <Scene key="_Launch" component={Launch}/>
</Scene>

```
look like mac os Launchpad

### DIFF
--- a/src/DefaultRenderer.js
+++ b/src/DefaultRenderer.js
@@ -211,6 +211,11 @@ export default class DefaultRenderer extends Component {
     }
 
     if (SceneComponent) {
+      if (navigationState.tabs || ! navigationState.parent) {
+        return (
+          <SceneComponent {...this.props} {...navigationState} />
+        );
+      }
       return (
         <View
           style={[styles.sceneStyle, navigationState.sceneStyle]}
--- a/src/DefaultRenderer.js
+++ b/src/DefaultRenderer.js
@@ -211,11 +211,6 @@ export default class DefaultRenderer extends Component {
     }
 
     if (SceneComponent) {
-      if (navigationState.tabs || !navigationState.parent) {
-        return (
-          <SceneComponent {...this.props} {...navigationState} />
-        );
-      }
       return (
         <View
           style={[styles.sceneStyle, navigationState.sceneStyle]}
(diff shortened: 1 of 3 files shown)


joenoon:
Comment from other PR: https://github.com/aksonov/react-native-router-flux/pull/827#issuecomment-226388281

src/DefaultRenderer.js
```diff
@@ -211,6 +211,11 @@ export default class DefaultRenderer extends Component {
     }
 
     if (SceneComponent) {
+      if (navigationState.tabs || !navigationState.parent) {
```
joenoon: What is the need for having a different code path here?  Is there something in the Example app I can check?  Does this change break any of the Example app pages?

phproot: sorry,my bad.

phproot: if sceneStyle set paddingTop、alignItems... 
modal、tabs、NavigationRootView will inherit these attributes。

joenoon: @phproot no problem! you can probably use getSceneStyle to account for the hideNavBar/hideTabBar props like in the Example app, and not hard-code a marginTop in sceneStyle.  Check the Example app that does exactly this.

That 3D screenshot is great.  Is that Reveal.app or something else?

joenoon:
@aksonov this seems ok to me, except I'm not sure if there could be unexpected problems with the `condition`.  That whole part of Actions is a little unclear to me.

src/DefaultRenderer.js
phproot: @joenoon is xcode 
Debug -> View Debugging -> Capture View Hierarchy

aksonov:
@alextkd Is it OK to merge?

phproot:
What are Can we merge?

alextkd:
Looks good to me. I don't see any problems it can cause.

PR #831, opened 2016-06-16
Fix typescript definition

xirc:
Fix TypeScript definition.

- Use 'react-native' module instead of 'react' one to reference types which is defined by 'react-native'.  
  It seems to be proper way although 'react-native's typing extends namespace __React.
- Fix RNRFActions.create definition
- Remove duplicate declaration of RouterProps.getSceneStyle

I checked that the definition works fine at
"react-native": "registry:dt/react-native#0.19.0+20160501140141".

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -269,7 +269,7 @@ declare namespace RNRF {
     jump(props: props): void,
     refresh(props: props): void,
     focus(props: props): void,
-    create(scene: Scene, wrapBy?: () => any): Object
+    create(scene: React.ReactNode, wrapBy?: () => any): Object
   }
 
   export var Actions: RNRFActions;
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,6 +2,8 @@
 // created by bang88 (https://github.com/bang88)
 
 import * as React from 'react';
+import * as ReactNative from 'react-native';
+
 
 declare namespace RNRF {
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -269,7 +269,7 @@ declare namespace RNRF {
     jump(props: props): void,
     refresh(props: props): void,
     focus(props: props): void,
-    create(scene: Scene, wrapBy?: () => any): Object
+    create(scene: React.ReactNode, wrapBy?: () => any): Object
```
bang88: Why this ?

xirc:
Sorry, I had a mistake to define Actions.create.
a545cad makes you to be able to write the following code.
```
Actions.create(
  <Scene key={"some_key"} />
);
```

index.d.ts
xirc: Actually, I want to define
```
create(scene: Scene, wrapBy?: () => any): Object
```
But, using this definition, I cannot write the following code.
```
Actions.create(
  <Scene key={"some_key"} />
);
```

Are there better way?

bang88: `scene `here is extends from `React.Component` it should works .By the way. are you using `typescript` with `react-native`.If so can you provide an online version,I can have a look.

xirc: https://github.com/xirc/react-native-router-flux-sandbox is the example.
'master' branch uses 69e0ab4
'fixed-one' branch uses a545cad

bang88: I see,you are right.Thanks

xirc: No problem. Thanks 😄

PR #835, opened 2016-06-17
Revert "Feature Hand Back Key for Android"

aksonov:
Reverts aksonov/react-native-router-flux#820


yanqiw:
I suggest that only throw the error when platform is 'android' and 'state.index===0', and let developer to handle the android back action in application.

PR #843, opened 2016-06-21
Prefix action strings to namespace them.

ThaJay:
I noticed the action constants are very generic strings (not namespaced like 'router_jump' etc.):
```
export const JUMP_ACTION = 'jump';
export const PUSH_ACTION = 'push';
export const REPLACE_ACTION = 'replace';
export const POP_ACTION2 = 'back';
export const POP_ACTION = 'BackAction';
export const REFRESH_ACTION = 'refresh';
export const RESET_ACTION = 'reset';
export const FOCUS_ACTION = 'focus';
```
As I understand Redux, all actions get passed to all reducers when you use `combineReducers`. We see these strings as being so generic that it opens a very real risk of me (or another module I did not read) dispatching an action that evaluates to the same string, triggering an action that was not supposed to be triggered. To prevent this I propose the following change, or any other that diminishes this risk.

At the moment this is not an issue with Redux as we are unable to combine both stores (probably only possible with MobX at the moment?), but as RNRF becomes store agnostic we would need something like this.

### DIFF
--- a/src/Actions.js
+++ b/src/Actions.js
@@ -8,15 +8,15 @@
  */
 import { assert } from './Util';
 import Scene from './Scene';
-export const JUMP_ACTION = 'router_jump';
-export const PUSH_ACTION = 'router_push';
-export const REPLACE_ACTION = 'router_replace';
-export const POP_ACTION2 = 'router_back';
-export const POP_ACTION = 'router_BackAction';
-export const POP_TO = 'router_popTo';
-export const REFRESH_ACTION = 'router_refresh';
-export const RESET_ACTION = 'router_reset';
-export const FOCUS_ACTION = 'router_focus';
+export const JUMP_ACTION = 'REACT_NATIVE_ROUTER_FLUX_JUMP';
+export const PUSH_ACTION = 'REACT_NATIVE_ROUTER_FLUX_PUSH';
+export const REPLACE_ACTION = 'REACT_NATIVE_ROUTER_FLUX_REPLACE';
+export const POP_ACTION2 = 'REACT_NATIVE_ROUTER_FLUX_BACK';
+export const POP_ACTION = 'REACT_NATIVE_ROUTER_FLUX_BACK_ACTION';
+export const POP_TO = 'REACT_NATIVE_ROUTER_FLUX_POP_TO';
+export const REFRESH_ACTION = 'REACT_NATIVE_ROUTER_FLUX_REFRESH';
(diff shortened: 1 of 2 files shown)


charpeni:
CircleCI failed.

aksonov:
@joenoon @charpeni Are you ok with this change?

joenoon:
@aksonov I dont use redux and haven't had a need for this.  I guess if its really necessary then its fine... but would be good to have someone else verify its a real problem that needs solving.

aksonov:
@rturk Are you using redux?

aksonov:
cc @cridenour

cridenour:
Not using redux so I can't say for certain. I had the same understand though that all redux actions are combined, but maybe this is something we can accomplish in the redux example in the docs?

charpeni:
I can see the use case, but i'm not a redux expert.

zxcpoiu:
It's true.
All reducers will be invoked if it filters the same action name.
It's better than nothing in my opinion.

Common approach is to give the action name as less conflict as possible, like ROUTER_FLUX_ACTION_JUMP and so on

zxcpoiu:
And it seems we can combine store if you follow the instruction here:
https://github.com/aksonov/react-native-router-flux/blob/master/docs/REDUX_FLUX.md

cridenour:
I just think this is supposed to be a flux agnostic library, so while redux would like more specific action names, can we do this on the fly as part of the optional redux docs? If so, I would prefer that.

zxcpoiu:
Yeah, both works. Either change it or as optional config.

Since the module is flux agnostic, it is worthy to prefix action name with full module name which makes it clearer actions and avoid potential conflict no matter what flux implementation you are using.

how about REACT_NATIVE_ROUTER_FLUX_JUMP ?
It won't hurt as long as we can import action name as constant.

joenoon:
Looks good

aksonov:
@ThaJay Could you also update Example and docs to avoid complaints #894 , thanks..

zxcpoiu:
I can do that. wait a minute

aksonov:
@zxcpoiu We need to rework it without breaking existing apps.

aksonov:
Like create map (short names -> redux names) and use it only with `dispatch` call

zxcpoiu:
got it.

aksonov:
new PR will be welcome.

farwayer:
It should have been minor version bump not patch...

ThaJay:
Sorry, haven't checked here in a few days.
Thanks @zxcpoiu for picking it up in the meantime

PR #846, opened 2016-06-22
Replace `assert` in Switch.js with custom `assert`

Kerumen:
It breaks the component because `assert` is no longer in the dependencies..

### DIFF
--- a/src/Switch.js
+++ b/src/Switch.js
@@ -1,6 +1,6 @@
 import React, { PropTypes } from 'react';
 import TabBar from './TabBar';
-import assert from 'assert';
+import { assert } from './Util';
 
 export default function Switch(props) {
   const navState = props.navigationState;


aksonov:
Ops, thanks.

teyou:
finally... i just bumped into this issue 30 mins ago after upgraded the npm package

Kerumen:
> finally...

The fix took less than 1 hour.

aksonov:
We definitely need to have unit tests for our views to avoid such problems... I'm having assert within my app, so i didn't mention that issue at all.

Kerumen:
Don't forget to `npm prune` frequently.
And when you publish a new version `rm -rf node_modules && npm i` and a compilation should be done to start from a clean install to check these errors.

charpeni:
> We definitely need to have unit tests for our views to avoid such problems... I'm having assert within my app, so i didn't mention that issue at all.

@aksonov I'm working on something.

PR #851, opened 2016-06-23
Use ./components/TabIcon

sarovin:


sarovin:
Now you can see an error with TabIcon:

> Warning: Failed propType: Invalid prop `selected` of type `boolean` supplied to `TabIcon`, expected `string`. Check the render method of `Tabs`.

PR #861, opened 2016-06-25
Fix missing brackets in docs

dustin-H:


charpeni:
Good catch, thank you!

PR #868, opened 2016-06-28
Allow for one scene to fade into the next in addition to horizontal/vertical

andyschwob:
This would allow you to use ```direction="fade"``` as another option to vertical when transitioning from one scene to another.  This could be a fix until something like this is added to NavigationExperimental.

I tried to accomplish this without the need for changes to DefaultRenderer by passing in animationStyle, but I couldn't figure out how to access the NavigationSceneRendererProps (location, position, scene) from Example.js, or if it's even possible.

### DIFF
--- a/docs/API_CONFIGURATION.md
+++ b/docs/API_CONFIGURATION.md
@@ -37,6 +37,8 @@
 |-----------|--------|---------|--------------------------------------------|
 | duration | `number` | | optional. acts as a shortcut to writing an `applyAnimation` function with `Animated.timing` for a given duration (in ms). |
 | direction | `string` | 'horizontal' | direction of animation horizontal/vertical |
+| animation | `string` | | animation options when transitioning: 'fade' currently only option |
+| animationStyle | `function` | | optional interpolation function for scene transitions: `animationStyle={interpolationFunction}` |
 | applyAnimation | `function` | | optional if provided overrides the default spring animation |
 
 ### Scene styles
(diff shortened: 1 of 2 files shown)


aksonov:
Looks nice, please update docs with new possible value for direction.

joenoon:
This looks good to me - one other way to do this might be to add `animationStyleFn` or do some type checking on `animationStyle` and if its a fn then call it with the sceneprops.  That would allow stuff like this and anything else.  But its probably nice to have some included transitions using `direction` like this PR.

aksonov:
@andyschwob  btw, 'direction' looks not suitable name for this, could we use something like 'animation' (and support old 'direction' as well). Also as @joenoon mentioned, it would be great to do some generalization function like animationStyle or something like that...

andyschwob:
I agree that direction is not a suitable name, I'll change it to support both.  Good idea with the type checking to call the function, I'll try to add that in too.

andyschwob:
@aksonov @joenoon 
updated DefaultRenderer and docs to include a couple changes:
- You can now pass in animation="fade" to fade into the next scene.  If a direction is also specified, the direction will take precedence over the fade, if none are specified it defaults to horizontal.
- You can instead pass your own interpolation function into the scene with animationStyle like: ```animationStyle={customInterpolationFunction}```

aksonov:
@andyschwob Thanks!

SSTPIERRE2:
When you say v3 is obsolete now, do you mean that community built features like the fade animation no longer work/apply to v4?

With the latest version applying the prop "animation='fade'" doesn't work for me/my team.

aksonov:
@SSTPIERRE2 You can still use v3, I just cannot support it (because of focus on v4). There is ticket about adding animations to v4: #2042

PR #870, opened 2016-06-28
Add parent's NavBar style for TabBar component

sarovin:
i can't use navigationBarStyle props for set a custom navbar for all tabs.
Example:

```
<Scene
  key="main"
  navigationBarStyle={{ backgroundColor: 'white' }} // <-- This Not Work on tabs
  tabs
  tabBarStyle={styles.tabBarStyle}
  tabBarSelectedItemStyle={styles.tabBarSelectedItemStyle}
>
  <Scene
    key="tab1"
    title="Tab #1"
    icon={TabIcon}
    navigationBarStyle={{ backgroundColor: 'red' }}
    titleStyle={{ color: 'white' }}
    component={TabView}
  />
  <Scene key="tab2" component={TabView} initial title="Tab #2" icon={TabIcon} />
// other tabs
</Scene>
```

### DIFF
--- a/src/TabbedView.js
+++ b/src/TabbedView.js
@@ -54,9 +54,10 @@ class TabbedView extends Component {
         return;
       }
       this.renderedSceneKeys[key] = true;
-      item.navigationBarStyle = this.props.navigationState.navigationBarStyle;
-      item.titleStyle = this.props.navigationState.titleStyle;
-      scenes.push(this.renderScene(item, i));
+      const child = item;
+      child.navigationBarStyle = this.props.navigationState.navigationBarStyle;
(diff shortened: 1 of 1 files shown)


cridenour:
Doesn't this overwrite any styles passed to the child intentionally?

sarovin:
@cridenour Yes, it not overwrite any styles...
@joenoon have a better solution

joenoon:
@sarovin this broke some things for others so I think you might need to provide your own navBar or see if there is another way to just pass the styles everywhere needed in the jsx.

aksonov:
@sarovin custom navbar is really correct way to go here.

PR #875, opened 2016-06-28
Added optional tabBarIconContainerStyle property to override the style of the icon container view.

SudoPlz:
Had trouble styling the tab icons the way I wanted, so I exposed the iconStyle property of the TabBar the the scene component, under the name tabBarIconContainerStyle.

Also added it to the documentation.


charpeni:
LGTM.

PR #886, opened 2016-06-30
refactor #870

joenoon:
refs #870 

seems like a safe way to go? lets the item still have last say

### DIFF
--- a/src/TabbedView.js
+++ b/src/TabbedView.js
@@ -54,7 +54,8 @@ class TabbedView extends Component {
         return;
       }
       this.renderedSceneKeys[key] = true;
-      scenes.push(this.renderScene(item, i));
+      const {navigationBarStyle,titleStyle} = this.props.navigationState;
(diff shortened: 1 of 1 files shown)


src/TabbedView.js
```diff
@@ -54,7 +54,8 @@ class TabbedView extends Component {
         return;
       }
       this.renderedSceneKeys[key] = true;
-      scenes.push(this.renderScene(item, i));
+      const {navigationBarStyle,titleStyle} = this.props.navigationState;
+      scenes.push(this.renderScene({navigationBarStyle,titleStyle,...item}, i));
     });
     return (
       <View style={this.props.style}>
```
aksonov: as i've found today this commit broke my app :(

joenoon: @aksonov sorry about that! what did it break?

aksonov: Right button style and overall style for own nav bar passed to Router globally

Pavel.

Issue #894, opened 2016-07-02
Scene {type:'reset'} or type="reset" is not working in 3.30.4 version.


Durgaprasad-Budhwani:
### Version
Tell us which versions you are using: 
Recently I migrated to react-native-router-flux v3.30.4 version from react-native-router-flux v3.30.3.
Scene which has  prop '**type**' is not navigating. 

- react-native-router-flux v3.30.4
- react-native v0.28.0

### Expected behaviour
Scene with **type** prop is working fine.

### Actual behaviour
Not able to navigate to scene which has prop **type**

cridenour:
Is this something you're also seeing in the Example app?

Durgaprasad-Budhwani:
@cridenour Yes same thing is also reproducible in Example app and to reproduce there, go to register page in example app and click on replace page or go to switch page and click on exit, it will not work.

felipemartim:
also happening here

iWinkey:
Same issue with type="replace"

iWinkey:
2fff2c9efd410bde2839c894f5e91f2f3b7ae8c6 See this commit and #843 this discussion!
Now the type string is prefixed to namespace them, so the type="reset" now is type="REACT_NATIVE_ROUTER_FLUX_RESET" and type="replace" now is type="REACT_NATIVE_ROUTER_FLUX_REPLACE"

ElvisChiang:
FYI,

change from 
```JavaScript```
import { Scene, Reducer, Router } from 'react-native-router-flux';

<Scene key={deviceConfg} type="replace' >
```
to 
```JavaScript
import { Scene, Reducer, Router, Actions } from 'react-native-router-flux';
const { PUSH_ACTION, REPLACE_ACTION } = Actions;

<Scene key={deviceConfg} type={REPLACE_ACTION} >
```

cridenour:
Thank you @iWinkey. This is an unfortunate and unintentional change of a merge request aimed at better redux integration. It might be worth rolling back, but I'll wait for others to chime in.

Durgaprasad-Budhwani:
Thank you @iWinkey. It is working like a charm. Closing this issue.

PR #900, opened 2016-07-05
normalize action name with prefix and expose it from module

zxcpoiu:
This is a follow up of #843 which has been reverted (#899) since it breaks existig apps as described in #894.

I have tested on Example follow the [steps to reproduce](https://github.com/aksonov/react-native-router-flux/issues/894#issuecomment-230142299) provided by @Durgaprasad-Budhwani


zxcpoiu:
@aksonov is these what you mean?

If it is ok, I will add commits to change documents accrodingly.

src/Router.js
```diff
@@ -88,6 +88,9 @@ class Router extends Component {
     }
 
     Actions.callback = props => {
+      if (props.type && sceneTypeMapAction[props.type]) {
+        props.type = sceneTypeMapAction[props.type];
+      }
       if (this.props.dispatch) this.props.dispatch(props);
```
aksonov: My idea was slightly different - don't modify type (to avoid breaking existing apps), but just pass modified type to 'dispatch' method only

aksonov:
Almost, check my comment.

zxcpoiu:
* do you mean something like this?

```javascript
if (this.props.dispatch) this.props.dispatch({...props, type: sceneTypeMapAction[props.type]});
```

* so I don't need to pass modified type to `onNavigate`, right?

cridenour:
What does this do to the router configs with type="replace" and type="reset"? This will also need to update the Example app.

zxcpoiu:
@cridenour 
thanks for review!
and yes, I will add Example and related documents these days.

basically, this would not do anything about that as @aksonov suggested not to modified original props to avoid issues which described in #894 

in short, in `Router.js`, if and only if  `this.dispatch()` exist in props, this PR will translate actions to unique action something like from `'reset'` to `'REACT_NATIVE_ROUTER_FLUX_RESET'`.

so it should not break existing apps.

if you found any chances that it would potentially breaks something, please let me know. 


only for app which is using redux-like store needs to use action constants imported from `ActionConstants.js`. like:

```javascript
import { ActionConst } from 'react-native-router-flux'; // <===== CHANGE

const initialState = {
  scene: {},
};

export default function reducer(state = initialState, action = {}) {
  switch (action.type) {
    // focus action is dispatched when a new screen comes into focus
    case ActionConst.FOCUS_ACTION: // <===== CHANGE
      return {
        ...state,
        scene: action.scene,
      };

    // ...other actions

    default:
      return state;
  }
}
```

and you can also use constants for consistency if you want. ( it's better IMO )
ex:

```javascript
Actions.ROUTE_NAME({type: ActionConst.RESET_ACTION});

or 

<Scene key="myscene" type={ActionConst.REPLACE_ACTION} >
```

zxcpoiu:
by the way, you reminded me that, do you think we should strip all `_ACTION` suffix in the constant name?

since we can import constants from everywhere now, may be it would be more clearer if we use:

```javascript
export const JUMP = 'REACT_NATIVE_ROUTER_FLUX_JUMP';
export const PUSH = 'REACT_NATIVE_ROUTER_FLUX_PUSH';
export const REPLACE = 'REACT_NATIVE_ROUTER_FLUX_REPLACE';
export const BACK = 'REACT_NATIVE_ROUTER_FLUX_BACK';
export const BACK_ACTION = 'REACT_NATIVE_ROUTER_FLUX_BACK_ACTION';
export const POP_TO = 'REACT_NATIVE_ROUTER_FLUX_POP_TO';
export const REFRESH = 'REACT_NATIVE_ROUTER_FLUX_REFRESH';
export const RESET = 'REACT_NATIVE_ROUTER_FLUX_RESET';
export const FOCUS = 'REACT_NATIVE_ROUTER_FLUX_FOCUS';

=== and ===

export const sceneTypeMapAction = {
  jump: ActionConst.JUMP,
  push: ActionConst.PUSH,
  replace: ActionConst.REPLACE,
  back: ActionConst.BACK,
  BackAction: ActionConst.BACK_ACTION,
  popTo: ActionConst.POP_TO,
  refresh: ActionConst.REFRESH,
  reset: ActionConst.RESET,
  focus: ActionConst.FOCUS,
};
``` 

then we can do this:

```
import { ActionConst } from 'react-native-router-flux';

Actions.ROUTE_NAME({type: ActionConst.RESET});

or 

<Scene key="myscene" type={ActionConst.REPLACE} >
```

thoughts?

cridenour:
Thanks @zxcpoiu! 

I think removing the _ACTION makes sense as we'll be using `ActionConst` in front of it in almost every situation.

zxcpoiu:
yeah, cool
Let's do it

zxcpoiu:
**change description:**

  * all actions now defined in a separate file: `ActionConst.js`.
  * create an mapping object called: `ActionMap` in `Actions.js`,
    it maps deprecated string literal actions to constant one.

**router:**

  * will dispatch constant actions if we have dispatch method in props
  * will pass unmodified type into onNavigate.

**reducer:**

  * always use ActionConst and ActionMap to determine whether an action type matched or not


**breaking**

  it's really hard to not break ANY existing app.
  especially for those who:
    1. stores routing status outside from RNRF
    2. use custom reducer instead of RNRF default reducer
  
  should be always use action constant from RNRF.


it will **NOT** break if user uses RNRF default reducer.

aksonov:
@zxcpoiu Sounds good, will you submit PR? We will publish this change under 3.31.0

zxcpoiu:
I want to make sure before I change document.

* we want to use constant actions across all module

* in `Actions.js`, it will pass **constant action** to it's callback;

* in `Router.js` current logic when defining `callback` is:
  only use constant action if we have `dispatch` method.
  but did not modify type which passed in `onNavigate()`

* Reducer will receive actions from `onNavigate`, and uses `ActionMap` to handle both constant and string literal actions to compare with `ActionConst`


since reducer uses ActionConst as well, so the only problem would be for those who are using custom reducer, should change action name in their code corresponds with `ActionConst`. ( no matter we pass original string literal action type or constant action type to `onNavigate` )

in that case, can I say:

when defining callback function in `Router.js`,
we should modified all actions type to ActionConst if the action name is belonged to `reservedKeys` ?

zxcpoiu:
@aksonov I already rebase my PR corresponds to my description.

But I an not sure if it is OK.
please see my comments above.

zxcpoiu:
looks at code is much clearer.

**my whole point is:**

since
`Actions.js` and `Reducer.js` always use action constant
and
for those who using custom reducer or redux store, must change their code to use constant as well.

should we change code in `Router.js` from:

```javascript
    Actions.callback = props => {
      if (this.props.dispatch) {
        if (props.type && ActionMap[props.type]) {
          this.props.dispatch({ ...props, type: ActionMap[props.type] });
        } else {
          this.props.dispatch(props);
        }
      }
      return onNavigate(props);
    };
```
to

```javascript
    Actions.callback = props => {
      if (props.type && ActionMap[props.type]) {
        props.type = ActionMap[props.type];
      }
      if (this.props.dispatch) {
          this.props.dispatch(props);
      }
      return onNavigate(props);
    };


```

aksonov:
Most users don't use any constants and custom reducers and the component should work for them.

zxcpoiu:
alright, I have changed to:

* if action type is existed in our reservedKeys, we will change it to constant value.
* we can handle both original string literal actions and action consts.  

have tested with Example.
will change Example and Document accordingly soon.

zxcpoiu:
done!

let me know if there are something need to be adjust

aksonov:
@zxcpoiu Great!

PR #915, opened 2016-07-10
throw error when pop action is called on root scene in android.

yanqiw:
This change is for handle the back action in android.  In android, when user click the back button, the pop() action is trigger, however,  we want to know the pop is triggered on root scene, to handle the exit app action. 
refer to [Feature Hand Back Key for Android](https://github.com/aksonov/react-native-router-flux/pull/820)

### DIFF
--- a/docs/ANDROID_BACK. md
+++ b/docs/ANDROID_BACK. md
@@ -1,0 +1,47 @@
+Android back action
+====
+This is an example to handle `back button action` in android.
+
+The reducer will throw an error when the back button is clicked on root scenes. You can catch the error, and handle the back action as you want.
+
+# Example
+
+`main.js`
+```Javascript
+import { BackAndroid } from 'react-native'
+
+backButtonPressedOnceToExit = false;
+
+export default class Main extends React.Component {
+...
+  componentWillMount(){
+    BackAndroid.addEventListener('hardwareBackPress', this.onBackAndroid.bind(this));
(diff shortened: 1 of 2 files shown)


yanqiw:
hi, dose anybody is checking this request?

cridenour:
Unfortunately I'm not building an Android app at the moment so I haven't have a chance to dive in. However, your assert doesn't seem to be in the right format. 

That said, if this is a well documented UX pattern, we should include the exit in our own code rather than a documentation suggestion. If it's not, I think pointing people to a custom reducer is best.

joenoon:
I'm also not sure about this because I'm not working on an Android app.  But a couple of notes:

1. The doc duplicates componentWillMount instead of having a componentWillUnmount with removeEventListener.
2. Something doesn't feel right with catching an exception as the official way to handle this case.
3. I haven't tested the reducer modification, but at a glance of the diff I wonder if this works on a nested navstate tree, where there are multiple levels of state with children, and multiple levels where index === 0.  Does it make sure its acting on the index of the true root?

yanqiw:
@joenoon @cridenour thx for the commons, i think u are right,  i will cancel this request, and find a better way to done this, thx.

yanqiw:
@joenoon @cridenour thx for the comments, i think u are right, i will cancel this request, and find a better way to done this, thx.

PR #916, opened 2016-07-10
Document PopNum

K-Leon:
I had a hard time figuring out that adding popNum to Actions.pop is possible. So i would love to see it mentioned somewhere in the documentation. I hope this is an appropriate place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ import {Actions} from 'react-native-router-flux'
 And then:
 
 * `Actions.ACTION_NAME(PARAMS)` will call the appropriate action and params will be passed to the scene.
-* `Actions.pop()` will pop the current screen.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,7 +108,7 @@ import {Actions} from 'react-native-router-flux'
 And then:
 
 * `Actions.ACTION_NAME(PARAMS)` will call the appropriate action and params will be passed to the scene.
-* `Actions.pop()` will pop the current screen.
+* `Actions.pop()` will pop the current screen. It takes also param {popNum: [number]} to pop multiple screens at once.
```
charpeni: I would probably write something like : It can also takes param [...]

K-Leon: fixed this - i hope. sorry for my bad english.

Issue #893, opened 2016-07-02
Improve documentation about how to implement modals


kennethlynne:
The current documentation is kinda "go fuck your self" for newbies just exploring this great router.

kennethlynne:
Just to be clear: I'm a newbie, or else I would send a PR 👍

JimmyStavros:
Agree. It also seems like the example is pretty outdated compared to the current API.

sondremare:
Trying to use modals as the example code shows. And it crashes the entire app when used with a view with https://github.com/FaridSafi/react-native-gifted-messenger

aksonov:
@sondremare Check Example project, does modal work for you?

brien-crean:
Agree, also a bit confused about Modal usage.
If using a Modal as a way of reporting App status to the user e.g. "Network connection failed" etc. 
displaying the Modal works fine I can show it on each screen as the user navigates. 
However as Actions.pop() (or maybe ActionConst.BACK in latest version) is used to dismiss the Modal, if I push another route to the stack and then try to dismiss the Modal it instead pops the most recently added route from the stack and the Modal is stuck there

Is there a better way to implement or is it possible to override onNavigate on the Modal component for example? I tried adding it as a prop to the Scene but it prevented the component from rendering
E.g.
```
      <Router createReducer={this.reducerCreate.bind(this)} >
        <Scene key="modal" component={Modal} >
          <Scene key="root">
            <Scene key="loginScreen" hideNavBar={true} initial={this.state.needSignIn} component={LoginScreen} type="reset" duration={200} />

          </Scene>
            <Scene key="statusModal" component={ErrorModal} onNavigate={ ()=> Actions.loginScreen()}/>
        </Scene>
      </Router>
```

heat:
I have some trouble with Actions.pop() when switch tabs while modal is open.

Is there a action to close modal or what @charpeni is the best way

brien-crean:
@heat @cridenour I tried to submit a pull request, but got permission denied. Here is the example of Modal usage I was trying to submit:

```jsx
import StatusModal from './components/StatusModal'

<Router>
  <Scene key="modal" component={Modal} >
    <Scene key="root">
      <Scene key="screen1" initial={true} component={Screen1} />
      <Scene key="screen2" component={Screen2} />
    </Scene>
      <Scene key="statusModal" component={StatusModal} />
  </Scene>
</Router>
```

[...]

PR #923, opened 2016-07-13
added Modal example usage to OTHER_INFO.md for issue number #893

brien-crean:
I added an example of how to utilize the Modal component and dismiss a Modal which avoids using Actions.pop(). Because using pop could potentially cause navigation issues if another route is pushed to the stack before the Modal is dismissed.

### DIFF
--- a/docs/OTHER_INFO.md
+++ b/docs/OTHER_INFO.md
@@ -11,7 +11,65 @@ This is a place for information that needs to be documented, but aren't long eno
 - Sub-Scenes
 
 ## Modals
-To display a modal use `Modal` as root renderer, so it will render the first element as `normal` scene and all others as popups (when they are pushed), see Example for more details.
+To display a modal use `Modal` as root renderer, so it will render the first element as `normal` scene and all others as popups (when they are pushed).
+For example:
+```jsx
+import StatusModal from './components/StatusModal'
+
+<Router>
+  <Scene key="modal" component={Modal} >
+    <Scene key="root">
+      <Scene key="screen1" initial={true} component={Screen1} />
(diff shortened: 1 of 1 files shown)


cridenour:
While I didn't test the code myself, this looks great! Thanks for the write up.

aksonov:
@brien-crean @cridenour Sorry, but i see extra complexity here as well as internal state is used (setState) together with navigationState (i believe it should be avoided). What is wrong with current Example approach - when Actions.pop() will just dismiss popup and Actions.POPUP_NAME will show it?

brien-crean:
@aksonov Yes Actions.POPUP_NAME will show and Actions.pop() will dismiss the Modal, as long as another route is not pushed onto the Stack **before** trying to dismiss the Modal. If that happens the Modal is no longer at the top of the stack so it is stuck there. React Native now has a built in Modal customizable component for pop ups.

I am using the Modal to display a status message that does not stop the user from navigating when it is displayed for example a network connectivity Modal: 

screenshot 2016-07-18 10 45 29

Is there a way to remove the Modal from the stack if it is no longer at the top?

The only other solution to this scenario that I could think of is to have the Modal as a component that can be wrapped around all other components and shown and hidden based on state.

aksonov:
@brien-crean it sounds reasonable. Ok, let it be as temp solution until we will find better way (maybe introduce removeScene(ACTION_NAME) method? What do you think?

brien-crean:
@aksonov yes I think that would be a great feature to have! I'm sure there are other scenarios where it would be very useful to be able to pull a scene from the stack. 

I did see that someone else was asking for a similar feature in this issue/question
#930

Issue #927, opened 2016-07-15
Background image


Roilan:
- react-native-router-flux v3.30.1
- react-native v0.29.0

I was looking over https://github.com/aksonov/react-native-router-flux/issues/205 and tried replicating what he did on the latest API and no luck getting a shared background image to render. Wondering if anyone has come across this or found a good way to do. I've provided my example Router component below.

https://gist.github.com/Roilan/c0058cd424389b9887483de5ca613255

The `Background` component, just has a `Image` with `children` being rendered inside of it.

Thanks!

cridenour:
Why not just render the router inside a view with a background image?

Roilan:
Thanks for your response @cridenour. Great suggestion and not sure why I didn't think of that before. So, I tried what you suggested and still having some issues with the transparent background. I provided a gist of how it looks now.

https://gist.github.com/Roilan/4e0942dceab4fd84897384729bffb40d

Seems like even with a transparent background it's still overlaying on top of the image.

cridenour:
Does your home component have a transparent background set as well? I believe most react-native components default to non-transparent.

Roilan:
Yeah, I've tried setting the `home` component background to `transparent`. No luck either. If I set it to another color, it takes over the whole background, it makes me believe that `Background` is just laying behind everything even when we set the background. Thoughts?

brien-crean:
@Roilan I tried something similar and could see that the `Background` component was there in the component tree, by using the Nuclide React Native Inspector but I could not get it to render. I tried playing around with the new zIndex for iOS also but nothing I did seemed to help

Roilan:
@brien-crean Thanks for looking into it. I'm having the exact same issue. Not sure what else to try. Maybe it's worth looking into add the option for a background component.

Roilan:
Found the issue. So, it's related to this https://github.com/aksonov/react-native-router-flux/issues/448 the `backgroundColor: transparent` was removed. Manually setting the background to transparent in the `NavigationCard` works but throws errors due to shadows. @cridenour @aksonov any thoughts on if it's even possible to do this?

Roilan:
Quick update, I found the issue. The deep inside the `NavigationCard` component, it has a `shadowColor` and `backgroundColor`. Setting both to `transparent` fixes it. I'll be making a PR shortly for optional style in the `NavigationCard` to override this.

Roilan:
[...]

PR #931, opened 2016-07-16
add prop for navigation card style

Roilan:
Adds a prop for adding / changing current styles in the navigation card. This is mainly for shared background images across scenes and solves https://github.com/aksonov/react-native-router-flux/issues/927.

Example:
`<Scene key='home' component={Home} title='Home' navigationCardStyle={{ backgroundColor: 'transparent', shadowColor: 'transparent' }} />
          `

### DIFF
--- a/src/DefaultRenderer.js
+++ b/src/DefaultRenderer.js
@@ -128,7 +128,7 @@ export default class DefaultRenderer extends Component {
   }
 
   renderCard(/* NavigationSceneRendererProps */ props) {
-    const { key, direction, animation, getSceneStyle } = props.scene.navigationState;
+    const { key, direction, animation, getSceneStyle, navigationCardStyle } = props.scene.navigationState;
     let { panHandlers, animationStyle } = props.scene.navigationState;
 
     const state = props.navigationState;
@@ -168,7 +168,7 @@ export default class DefaultRenderer extends Component {
       <NavigationCard
         {...props}
         key={`card_${key}`}
-        style={[animationStyle, style]}
+        style={[animationStyle, style, navigationCardStyle]}
         panHandlers={panHandlers}
         renderScene={this.renderScene}
       />


joenoon:
I think this and https://github.com/aksonov/react-native-router-flux/issues/927 can be solved with `getSceneStyle` as in the example: https://github.com/aksonov/react-native-router-flux/blob/822e83f08f5063cd33bc0c92030311c9acafb2af/Example/Example.js#L107

Could you also try the static `sceneStyle` param by passing what you are currently passing to `navigationCardStyle`?

If neither of those work let me know and I'll take a closer look.  I have a feeling `getSceneStyle` is what you need.

Roilan:
@joenoon Can't believe I missed that. Thanks for pointing out `getSceneStyle`, works great and exactly what I wanted.

brien-crean:
@joenoon I am trying to do something similar, but instead I am trying to put a Component in the foreground of every Scene, to serve as a status error message that can be displayed or hidden based on props on every scene, e.g. no network connection. 

The reason I am trying this is because the current Modal method usage causes Navigation issues if a route is pushed to the stack before the Modal is dismissed.

So I wrap each scene in an Error Modal as @Roilan showed me like this:

```jsx
// Create Component Wrapper
const ComponentWrapper = ComposedComponent => class extends Component {
  render() {
    return (
      <ErrorModal>
        <ComposedComponent {...this.props} />
      </ErrorModal>
    )
  }
}
```
```
// Inside the Router wrap the Component with the wrapper
<Router createReducer={this.reducerCreate.bind(this)} getSceneStyle={getSceneStyle} >
    <Scene key="root">
          <Scene key="loginScreen" hideNavBar component={ComponentWrapper(LoginScreen)}/>
    </Scene>
</Router>
```

```
const getSceneStyle = (props, computedProps) => {
  const style = {
    backgroundColor: 'transparent',
    shadowColor: 'transparent',
  };
  return style;
};
```


screenshot 2016-07-16 00 25 48

However all I get is the Modal on top of a white background. I'm not sure what I did wrong? Thanks

joenoon:
@brien-crean I think it would depend a lot of what the styling of ErrorModal is.  You probably need to do some manual debugging with different colors with some alpha so you can see how things are overlapping.  Or try the RN style inspector in-app and see if that helps.  And verify your ErrorModal implementation is rendering its child, etc...  really tough to say without seeing quite a bit more about the specifics.

brien-crean:
@joenoon Thanks, I'll try a simpler implementation, maybe I goofed on the styling

Roilan:
Hey @joenoon,

Wondering if you have any ideas on how to handle the previous route being showed behind the current route but transparent.

So, `home` pushes to `login` but the `home` component UI stays behind but slightly faded.

Issue #857, opened 2016-06-24
How to reset every Scene?


ms88privat:
### Version

Tell us which versions you are using: 
- react-native-router-flux v3.30.0
- react-native v0.26.2

Is it possible to reset the complete state of the router?
I want that my components of the scenes are mounting again.

The use case is, that I want to reset it after a user is logging out.

ms88privat:
I often do things like "if componentDidMount and data is not available, load something". But this doesn't work anymore, because I reset my Redux-Store (so no data available) but the components are not going to mount again because the Router still exist.

Maybe there should be an option on the new switch component if it should unmount the other case

cridenour:
When you're calling the page after logging out use `Actions.KEY({type: "reset"})`. We use this for logging out without issue.

ms88privat:
@cridenour Thanks, but this doesn't work for me. Tried to reset every scene I got, but my last selected Tab is still active and will not mount again.

rturk:
@cridenour I have a global wrapper that completely un-mounts my hole app when the user logs out, including redux state cleanup.
I like it since its a simple and elegant solution, with extra benefit that its smart to release used resources. 

The F8 App was my inspiration:
https://github.com/fbsamples/f8app/blob/master/js/F8App.js#L80

ms88privat:
@rturk Thanks for the hint again. Now I gave this solution a second thought and it should work with two Routers too.

Because in my case a simple Login Screen isn't good enough, I need routing there too. That's why it is integrated in the current Router setup.

rturk:
@ms88privat in this case I recommend you to use redux (or other state management framework of your preference). 
What I don't recommend is use two (or more!) routers.. this can be inefficient (from a memory and CPU perspective)..

ms88privat:
@rturk I'm using redux. But this is not the problem. What I need is something like this:

```jsx

render() {
  return this.props.authenticated ? this.renderHome() : this.renderLogin();
}

renderHome() {
 return <Router1 />;
}

renderLogin() {
  return <Router2 />;
}
```

I can't use the Switch Component of Router because it will keep both Scenes in Memory (which will prevent the componentDidMount Events from happening again also)

Any solution for this Problem? I think I have to use the pure Navigator for the second Router instead of the Router from Router-Flux.

cridenour:
AFAIK, if you navigate to a page via `type=reset`, then any component added to the stack after that will have its componentDidMount action called.

[...]

PR #932, opened 2016-07-17
added new switch feature - unmountScenes

ms88privat:
see https://github.com/aksonov/react-native-router-flux/issues/857


rturk:
@ms88privat this is interesting. Have you tested this with nested scenes?

ms88privat:
@rturk my Router is very nested and it seems to work fine.

I got three tabs in my Switch Scene. One of them is a normal Tab Scene. All working fine. There shouldn't any breaking changes, because it all depends on the new prop "unmountScene" which you have to add to your switch scene.

joenoon:
I'm using a much older version of the Switch component due to another bug I found but could never understand.  So probably best if @aksonov and others who are familiar with the current Switch take a good look at this.

aksonov:
LGTM

PR #936, opened 2016-07-18
fix(navigationState): avoid undefined key for root scenes

blackxored:
This is one of the issues I've been investigating within my app.
Apparently #911 broke scene key generation when the scene has no parent.

Before:

```javascript
Object {key: "undefined_undefined_0_home_", name: "home", sceneKey: "home_", parent: "home", type: "REACT_NATIVE_ROUTER_FLUX_PUSH"…}
```

After:

```javascript
Object {key: "0_home_", name: "home", sceneKey: "home_", parent: "home", type: "REACT_NATIVE_ROUTER_FLUX_PUSH"…}
```

### DIFF
--- a/src/State.js
+++ b/src/State.js
@@ -30,6 +30,12 @@ function getStateFromScenes(route, scenes, props) {
   return result;
 }
 
+function getSceneKey(parent, key, position, sceneKey) {
+  return [parent, key, position, sceneKey]
+    .filter(v => typeof(v) !== 'undefined' && v !== null)
+    .join('_');
+}
+
 export function getInitialState(
   route: {string: any},
   scenes: {string: any},
@@ -42,7 +48,7 @@ export function getInitialState(
     return {
       ...scenes.rootProps,
       ...route,
-      key: `${parent}_${key}_${position}_${route.sceneKey}`,
+      key: getSceneKey(parent, key, position, route.sceneKey),
       ...parentProps,
       ...getStateFromScenes(route, scenes, props),
     };


joenoon:
I was aware of the keys looking like this when parent or key was undefined.  But did it break something, or is it just a visual issue?  IMO those generated keys should be opaque.  But if its actually breaking something I'd like to understand it.

blackxored:
For one I'd say it breaks navigationState as part of DefaultRenderer#componentWillReceiveProps as the comparison will fail despite the fact there should be the same keys. I'm also investigating multiple FOCUS actions being triggered when Scene has no component set, and this might be affecting that as well. Let me know what you think.

joenoon:
So far I'm not seeing anything but I'm also not sure exactly what I'm looking for.  I added a line to show the key when `DefaultRenderer#componentWillReceiveProps` determines they are different:

```
console.log('dispatchFocusAction', nextProps.navigationState.key);
```

So far it logs the same keys and fires the same number of times with your code and original.

There is also a typo `getScenekey` vs `getSceneKey` and an eslint issue.

Do you have a case I could try with the example app or is this still in the investigation stage?

blackxored:
Thanks for spotting those, I've fixed the build. The case I'm testing against it's two Scenes (one at the root with Modal component and a child scene container with no component) and a tab scene container underneath that container.

joenoon:
Can you gist the test case or tell me how to reproduce?  What logging or checking are you doing to test?  I added my log line to the example and didn't spot an issue so far.  Can you reproduce in the example app?

blackxored:
Trying:
```jsx
<Scene key="root" component={Modal}>
  <Scene key="root" hideTabBar hideNavBar>
    <Scene key="main" tabs type="replace" initial>
      <Scene key="home" component={Some} title="Home"/>
    </Scene>
  </Scene>
</Scene>

aksonov:
Duplicate keys ‘root'?

blackxored:
That was a typo, sorry.

PR #947, opened 2016-07-20
Add pushOrPop action - it will pop if scene exists in the stack, push…

aksonov:
… otherwise


rseemann:
@aksonov how does one make use of this? I understand the concept of calling the action, but is there a type that I can add to the component to allow this to action to be called, instead of the default push?

PR #958, opened 2016-07-22
If Switch then Reset History Stack

sarovin:
Is there another solution for Reset History Stack of all children when Switch to another Scene with the selector??

### DIFF
--- a/src/Switch.js
+++ b/src/Switch.js
@@ -2,6 +2,23 @@ import React, { PropTypes } from 'react';
 import TabBar from './TabBar';
 import { assert } from './Util';
 
+function resetHistoryStack(child) {
+  const newChild = child;
+  newChild.index = 0;
+  for (let j = 0; j < child.children.length; j++) {
+    if (child.children[j].initial) {
+      newChild.index = j;
+      if (!child.tabs) {
+        newChild.children = [child.children[j]];
+      }
+    }
+    if (newChild.children[j].children) {
+      newChild.children[j] = resetHistoryStack(child.children[j]);
+    }
+  }
+  return newChild;
+}
+
 export default function Switch(props) {
   const navState = props.navigationState;
   const selector = props.selector;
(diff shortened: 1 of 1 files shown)


sarovin:
This fix my problem.

ms88privat:
@sarovin did you have a look at this https://github.com/aksonov/react-native-router-flux/pull/932 ?

maybe it fixes your problem too. It was merged only recently. Please try to set the prop "unmountScenes" on your switch component

sarovin:
yes, but not work for me...

ms88privat:
@sarovin you are right. If some scenes are pushed, my solution doesn't work like expected. I will try your solution now

ms88privat:
I tried your solution. It behaves slightly different then mine (starting from first Tab again, which is good), but you suffer from the same problem like my solution if a Tab has pushed Scenes.

sarovin:
Can you show me an example?

charpeni:
Note to the collaborator that will do the merge : Please use squash and merge.

ms88privat:
@sarovin I'm sorry, I have no time at the moment. It looks something like this:

- Scene as switch with case1 and case2
  - case1: login / loading
  - case2: tabs with tab1 and tab2
    - tab1: multiple scenes (PUSH a Scene here)
    - tab2: ...

Control Flow:
- switch from case1 -> case2 and then push a scene in tab1
- switch back from case2 -> case1
- login again from case1 -> case2
=> tab1 will have the pushed scene as initial scene (which breaks the whole navigation for the User at this point, because the scene is not pushed, it is the tab right now, without a backButton)

sarovin:
@ms88privat you use unmountScenes in selector component with my fix?

ms88privat:
@sarovin don't think so... but maybe. I will give it another try on 30.07. Right now I'm in the middle of a relocation.

sarovin:
I don't understand why "index !== navState.index" not return true at the second switch.

For example:

- Scene as switch with case1 and case2
  - case1: login
  - case2: tabs with tab1 and tab2
    - tab1: ...
    - tab2: button for logout 

Control Flow:

- switch from case1 -> case2 and then go to tab2 and logout
- switch back from case2 -> case1
- switch again from case1 -> case2

now "index !== navState.index" is false...why???? 😕

aksonov:
I believe it is bug-feature - for simplicity Switch is pure function now, but ideally it should call refresh to change index. But this way child components may be rendered twice... Any thoughts to avoid that? PR is welcome.

Pavel.

sarovin:
#996

PR #961, opened 2016-07-22
Scene - implemented getPanHandlers prop

philipshurpik:
implemented getPanHandlers prop to be able to use custom PanResponder

usage example:
const getPanHandlers = props => {
	// some code here
	return customPanResponder.createPanHandlers(props);
};


cridenour:
What do set as a default with getSceneStyle? Should we provide a default function (that includes the direction stuff) to help clean up some of that logic?

philipshurpik:
Probably it would be good to clean up logic by changing API to make everything similar, for example: 

- to have props: getSceneStyle, getPanHandlers, getAnimationStyle
- to remove animationStyle and panHandlers - that are currently mutable - it's not good and clear

It will change API - we need to add warnings and fallback 

I can make proposal PR for that, for example for 3.32 version, so we will have more clear API

What do you think about this?

philipshurpik:
I cleaned up code and make it pass eslint.

For future versions I think we can clean up API
But for current - it can be good improvement :)

philipshurpik:
Guys, any ideas about this pull request?    
Do we need to clean API, or remain API the same (plus add one more prop) ?

aksonov:
LGTM

PR #968, opened 2016-07-24
feat(navBar): add optional titleProps to NavBar

sondremare:
To be able to set properties like 'numberOfLines' and 'lineBreakMode' in the title, add a titleProps attribute for the navBar.


charpeni:
Thank you.

PR #974, opened 2016-07-26
feat(Actions): modify prototype to get scene props

enapupe:
I modified the prototype of `Actions[key]` in order to have access to the scene props. This is very useful in this auth scenario where you'd simple add a propType like `<Scene key="someSceneKey" forbidden={true} />` and have access just with `Actions.someSceneKey.forbidden`, etc

I don't know if prototypes are the best way to go but it was the only one to avoid breaking changes..

### DIFF
--- a/src/Actions.js
+++ b/src/Actions.js
@@ -156,20 +156,32 @@ class Actions {
       if (this[el.key]) {
         console.log(`Key ${el.key} is already defined!`);
       }
-      this[el.key] =
-        (props = {}) => {
-          assert(this.callback, 'Actions.callback is not defined!');
-          this.callback({ key: el.key, type: ActionConst.REFRESH, ...filterParam(props) });
-        };
+      const setup = (o) => {
+        const call =
+          (props = {}) => {
+            assert(this.callback, 'Actions.callback is not defined!');
+            this.callback({ key: el.key, type: ActionConst.REFRESH, ...filterParam(props) });
+          };
+        Object.setPrototypeOf(o, Object.getPrototypeOf(call));
+        Object.setPrototypeOf(call, o);
+        return call;
+      };
+      this[key] = setup(staticProps);
     }
     if (this[key]) {
       console.log(`Key ${key} is already defined!`);
     }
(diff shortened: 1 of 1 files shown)


aksonov:
@enapupe looks quite complex and it doesn't return changed props (with Actions.refresh), so we would receive complaints for it at the future for sure. Could you just re-use existing Actions.get(SCENE_KEY) (sorry, it is not documented yet, but it should return actual state of Scene with key==SCENE_KEY?

enapupe:
Yeah, I could definitely use `Actions.get()`, but i couldn't find anything like this. Could you link the source? Thanks for your attention.

aksonov:
Just try it :) It is set from Router like all other actions.

enapupe:
`ExceptionsManager.js:55 _reactNativeRouterFlux.Actions.get is not a
function`
There is no `get()` in the Actions.js file.. Could you be more specific?

enapupe:
`ExceptionsManager.js:55 _reactNativeRouterFlux.Actions.get is not a
function`
There is no `get()` in the Actions.js file.. Could you be more specific?

edit:
I see there is something in `Router.js:95` but I can't call it from `Actions`

enapupe:
Never mind, I was running outdated code while looking to master branch..

enapupe:
So, the function is there but it isn't finding my scene..  I don't see any reasons to not, also, the used function (`findElement`) has a lot more logic than a simple `find`. What is the role of `ActionConst.REFRESH` in this?

aksonov:
Strange, it worked well for (i'm using it to open/close drawer)

enapupe:
IDK why. My `navigationState` root scene has only ONE children (the first registered scene) but the scenes property has all the scenes there.. 

screen shot 2016-07-26 at 2 37 53 pm

enapupe:
@aksonov I was debugging that case where `get` does not return my scenes and realized the get only finds scenes in the current navigation stack, so it is definitely not a solution. I need a function to get props for ANY registered scene.

Issue #976, opened 2016-07-26
How to pop from cloned route to parent. 


kristojorg:
### Version
Tell us which versions you are using: 

- react-native-router-flux v3.32.0
- react-native v0.29.0

### Expected behaviour

Routes set up like this:
```
<Scene clone key="composer" component={Composer} hideNavBar hideTabBar />

<Scene
  key="myAccount"
  initial
  tabs
  hideNavBar
  hideTabBar={false}
  tabBarStyle={styles.tabBar}
  tabBarIconContainerStyle={styles.tabBarIconStyle}
  type={ActionConst.REPLACE}
>
  <Scene key="dashboard" component={Dashboard} tabTitle="Dashboard" initial icon={TabIcon} hideNavBar hideTabBar={false} />
  <Scene key="myStreamContainer" tabTitle="Stream" icon={TabIcon} hideNavBar hideTabBar={false}>
    <Scene key="myStream" component={Stream} hideNavBar hideTabBar={false} />
    {/* <Scene key="myComposer" component={Composer} hideNavBar hideTabBar />*/}
  </Scene>
  <Scene key="recorder" tabTitle="Record" icon={TabIcon} hideNavBar hideTabBar={false} >
    <Scene key="activityChooserView" component={ActivityChooserView} hideTabBar={false} initial hideNavBar />
    <Scene key="recorderView" component={Recorder} title="recorder" hideTabBar hideNavBar unmountScenes />
    <Scene key="userResponseView" component={UserResponseView} title="Response" hideTabBar hideNavBar unmountScenes />
  </Scene>
</Scene>

<Scene
  key="traineeAccount"
  tabs
  hideNavBar
  tabBarStyle={styles.tabBar}
  tabBarIconContainerStyle={styles.tabBarIconStyle}
  type={ActionConst.REPLACE}
>
  <Scene key="traineeStreamContainer" tabTitle="Stream" icon={TabIcon} hideNavBar hideTabBar={false}>
    <Scene key="traineeStream" component={Stream} hideNavBar hideTabBar={false} type={ActionConst.PUSH} />
    {/* <Scene key="traineeComposer" component={Composer} hideNavBar hideTabBar type={ActionConst.PUSH} />*/}
  </Scene>
</Scene>
```

[...]

Issue #952, opened 2016-07-21
Switch component don't work


sarovin:
### Version
Tell us which versions you are using: 

- react-native-router-flux v 3.31.2
- react-native v0.28.0

HI,
i have a problem with the Switch component.

i have this selector:

```
const selectorAuth = (user) => {
  if (user && (user.get('auth_token') || user.get('isGuest'))) {
    return 'main';
  }
  return 'auth';
};

```

and this Scenes:
```

<Scene key="modal" component={Modal}>
      <Scene
        key="root"
        hideNavBar
        navigationBarStyle={styles.navigationBarStyle}
        titleStyle={[Fonts.h1, styles.titleStyle]}
        component={connect(state => ({ user: state.user }))(Switch)}
        unmountScenes
        duration={1}
        selector={({ user }) => selectorAuth(user)}
        tabs
      >
        <Scene
          key="auth"
          rightTitle={'SKIP'}
          rightButtonTextStyle={[Fonts.button, styles.rightButtonTextStyle]}
          onRight={() => { props.dispatch(signInGuest()); }}
        >
          <Scene
            key="welcome"
            component={Welcome}
            title="Welcome"
            titleStyle={{ color: Colors.colorPrimary }}
            type="replace"
            initial
          />
          <Scene
            key="signin"
            component={SignIn}
            title="Accedi"
          />
          <Scene
            key="recoveryPass"
            component={RecoveryPass}
            title="Recupera Password"
          />
          <Scene
            key="signup"
            component={SignUp}
            title="Registrazione"
          />
          <Scene
            key="otpScreen"
            component={OTPScreen}
            title="Verifica il numero"
          />
          <Scene
            key="completeProfile"
            component={CompleteProfile}
            title="Completa Profilo"
          />
        </Scene>
        <Scene
          key="main"
          type="replace"
          tabBarStyle={styles.tabBarStyle}
          tabs
        >
          <Scene
            key="list"
            component={List}
            title={'Liste'}
            icon={BottomTabIcon}
          />
          <Scene
            key="localize"
            title={'Scegli il servizio'}
            icon={BottomTabIcon}
            tabs
            tabBarStyle={styles.tabBarSearchStyle}
            iconStyle={styles.iconSearchStyle}
            initial
          >
            <Scene
              key="delivery"
              component={Delivery}
              title={'Consegna'}
              icon={TopTabIcon}
            />
            <Scene
              key="pickup"
              component={Pickup}
              title={'Ritiro'}
[...]

PR #996, opened 2016-08-01
revert to old Switch-style, keeping unmountScenes.  needs review

joenoon:
refs https://github.com/aksonov/react-native-router-flux/issues/976
refs https://github.com/aksonov/react-native-router-flux/issues/952

### DIFF
--- a/src/Switch.js
+++ b/src/Switch.js
@@ -1,59 +1,71 @@
-import React, { PropTypes } from 'react';
-import TabBar from './TabBar';
-import { assert } from './Util';
+import React, { Component, PropTypes } from 'react';
 
-export default function Switch(props) {
-  const navState = props.navigationState;
-  const selector = props.selector;
-  const statem = props.statem;
-  if (!selector && !statem) console.error('Selector should be defined.');
-  let index = -1;
-  if (!selector) {
-    // support Statem - Harel statecharts machine!
-    navState.children.forEach((el, i) => {
-      assert(el.default || el.state,
-        `Either default or state should be defined for element=${el.key}`);
-      if (el.default) {
-        index = i;
-      } else {
-        if (el.state.active) {
-          index = i;
-        }
-      }
-    });
-    assert(index !== -1, 'No default scene is defined');
-  } else {
+import DefaultRenderer from './DefaultRenderer';
+import Actions from './Actions';
+
+export default class Switch extends Component {
+
+  static propTypes = {
+    navigationState: PropTypes.object,
+    onNavigate: PropTypes.func,
+    selector: PropTypes.func,
+    unmountScenes: PropTypes.bool,
+  };
+
+  constructor(props, context) {
+    super(props, context);
+    this.updateState = this.updateState.bind(this);
+    this.state = {};
+  }
+
+  componentDidMount() {
+    this.updateState(this.props);
+  }
+
+  componentWillReceiveProps(props) {
+    this.updateState(props);
+  }
+
+  updateState(props) {
+    const navState = props.navigationState;
+
+    const selector = props.selector;
+    if (!selector) console.error('Selector should be defined.');
+
     const selectedKey = selector(props);
     if (!selectedKey) console.error('Selector should return key.');
-    navState.children.forEach((el, i) => {
-      if (el.sceneKey === selectedKey) {
-        index = i;
-      }
-    });
-    if (index === -1) console.error(`A scene for key “${selectedKey}” does not exist.`);
-  }
-  let navigationState = index !== navState.index ? { ...navState, index } : navState;
 
-  if (props.unmountScenes) {
-    navigationState = {
-      ...navigationState,
-      children: [navState.children[navigationState.index]],
-      index: 0,
-    };
+    const selected = navState.children.filter(el => el.sceneKey === selectedKey);
+    if (!selected) console.error(`A scene for key “${selectedKey}” does not exist.`);
+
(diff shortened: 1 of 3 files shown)


joenoon:
Can someone verify unmountScenes works?  It may need to be moved before the `Actions[selectedKey]();`

src/Switch.js
```diff
@@ -1,59 +1,71 @@
...
+        ...navigationState,
+        children: [navState.children[navigationState.index]],
+        index: 0,
+      };
+    }
+
+    this.setState({ navigationState });
   }
-  let navigationState = index !== navState.index ? { ...navState, index } : navState;
-
```
zxcpoiu: I guess ` if (props.unmountScenes)` should wrapped in line 44
if selected scene != current scene ( scene changed), than we should unmount scene

zxcpoiu: or
use
`let navigationState = index !== navState.index ? { ...navState, index } : navState;`
to restore `navigationState`  back to `navState`

zxcpoiu:
I had tested in the following structure:

```jsx

// --- this got infinite loop
<Scene key="switch" initial={true} component={Switch} tabs={true} unmountScenes={true} selector={switchSelector}>
    <Scene key="login" initial={true}/>
    <Scene key="home"/>
</Scene>
```

and

```jsx
// --- this would not switch even selector returns 'home'

<Scene key="switch" initial={true} component={Switch} tabs={true} unmountScenes={true} selector={switchSelector}>
    <Scene key="login"/>
    <Scene key="home" initial={true}/>
</Scene>
```

joenoon:
@zxcpoiu thanks for checking! Just made a change to only do the unmountScenes logic when a switch will actually occur.

@ms88privat could you also verify?

sarovin:
@joenoon I can test it tomorrow

zxcpoiu:
I still got infinite loop in the first case with new commits.

It feels like a little weird to me, just want to make sure, is the following descriptions correct or not:

at the first case:
```jsx
// --- this got infinite loop
<Scene key="switch" initial={true} component={Switch} tabs={true} unmountScenes={true} selector={switchSelector}>
    <Scene key="login" initial={true}/>
    <Scene key="home"/>
</Scene>
```


* `navState`'s children is `login` and `home`, and initial index is '0' since `login` is the initial component
* that's assume in `selector`, we return 'home', then the selected `navigationState` is `home` and it's index mostly is 0 (or the index of it's initial child.)

so the reason I got infinite loop is because this `children: [navState.children[navigationState.index]],`
we always set navState to it's index-0 child which is `login`, but the selector return 'home' which is the index-1 child of navState and has it's own child indexes cache whcih is 0.

and no matter how, eventually, we've change the navState to selected child's navigationState
( I'm not sure, is this the right flow that we replace navState with child's state?)

zxcpoiu:
I have tried several ways, but with no luck so far.

If someone want to reproduce, here is test case:

```jsx
// --- this got infinite loop if switch selector returns 'home'

    <Scene key="root">
        <Scene key="switch" initial={true} component={Switch} tabs={true} unmountScenes={true} selector={switchSelector}>
            <Scene key="login" initial={true} type="replace" />
            <Scene key="home" tabs={true} type="replace">
                <Scene key="tab1" initial={true} />
                <Scene key="tab2" initial={true} />
            </Scene>
        </Scene>
    </Scene>
</Scene>
```

sarovin:
```
@@ -15,6 +15,7 @@ export default class Switch extends Component {
   constructor(props, context) {
     super(props, context);
     this.updateState = this.updateState.bind(this);
+    this.resetHistoryStack = this.resetHistoryStack.bind(this);
     this.state = {};
   }
 
@@ -26,6 +27,23 @@ export default class Switch extends Component {
     this.updateState(props);
   }
 
+  resetHistoryStack(child) {
+    const newChild = child;
+    newChild.index = 0;
+    for (let j = 0; j < child.children.length; j++) {
+      if (child.children[j].initial) {
+        newChild.index = j;
+        if (!child.tabs) {
+          newChild.children = [child.children[j]];
+        }
+      }
+      if (newChild.children[j].children) {
+        newChild.children[j] = this.resetHistoryStack(child.children[j]);
+      }
+    }
+    return newChild;
+  }
+
   updateState(props) {
     const navState = props.navigationState;
 
@@ -43,11 +61,8 @@ export default class Switch extends Component {
 
     if (navigationState.key !== navState.children[navState.index].key) {
       if (props.unmountScenes) {
-        navigationState = {
-          ...navigationState,
-          children: [navState.children[navigationState.index]],
-          index: 0,
-        };
+        navState.children[navState.index] = this.resetHistoryStack(navState.children[navState.index]);
+        navigationState = { ...navState };
       }
       Actions[selectedKey]();
     }
```

aksonov:
@sarovin Maybe i'm wrong, but i believe navigationState should be changed within Reducer, not within Switch. What do you think?

sarovin:
@aksonov you are right but i don't understand how work your Reducer  😕 
I try with reducer as soon as possible

aksonov:
Reducer works in the same way as Redux reducers. You may introduce param for JUMP action and add appropriate logic where JUMP is processed.

sarovin:
#1002

aksonov:
Closing it because of #1002, @joenoon does it work for you?

joenoon:
@aksonov so far so good!  Totally missed all the action in the other PR - nice work @sarovin and everyone else!

PR #1002, opened 2016-08-02
refactor #996

sarovin:


sarovin:
@aksonov Is it right?

aksonov:
@sarovin Yes, looks nice. Could you verify that it works for your scenarios? Example project?

sarovin:
It work fine for me.
My scenarios is here #952

zxcpoiu:
I got the same with the following example, still figuring it out.

Is `this.setState({ navigationState });` always supposed to be the selected `navigationState` ( `selected[0]` )?

```jsx
// --- this got infinite loop if switch selector returns 'home'

    <Scene key="root">
        <Scene key="switch" initial={true} component={Switch} tabs={true} unmountScenes={true} selector={switchSelector}>
            <Scene key="login" initial={true} type="replace" />
            <Scene key="home" tabs={true} type="replace">
                <Scene key="tab1" initial={true} />
                <Scene key="tab2" initial={true} />
            </Scene>
        </Scene>
    </Scene>
</Scene>
```

sarovin:
@zxcpoiu I don't understand your problem...i have tested with your scenarios and it work fine.

src/Switch.js
```diff
@@ -1,59 +1,64 @@
...
+
+    if (navigationState.key !== navState.children[navState.index].key) {
+      if (props.unmountScenes) {
+        Actions[selectedKey]({ unmountScenes: true });
       }
-    });
-    if (index === -1) console.error(`A scene for key “${selectedKey}” does not exist.`);
+      Actions[selectedKey]();
+    }
+    this.setState({ navigationState });
```
zxcpoiu: it will do `this.setState()` in the end using variable `navigationState` ( a child of `navState` )
so if the **second** time you are entering `this.updateState()`, navState would never match scene key returned by selector.

the first time you entered `this.updateState()`:  `navState.key` is 'switch', and it's children are ['login', 'home']
the second time you entered `this.updateState()`:  `navState.key` is 'home', and it's children are ['tab1', 'tab2']

my question: is `navState` supposed to be the child one eventually ?

src/Switch.js
```diff
@@ -1,59 +1,64 @@
...
     const selectedKey = selector(props);
     if (!selectedKey) console.error('Selector should return key.');
-    navState.children.forEach((el, i) => {
-      if (el.sceneKey === selectedKey) {
-        index = i;
+
+    const selected = navState.children.filter(el => el.sceneKey === selectedKey);
+    if (!selected) console.error(`A scene for key “${selectedKey}” does not exist.`);
+
+    const navigationState = selected[0];
```
zxcpoiu: `navigationState` is from `selected[0]`
so `navigationState` is a child of `navState`

src/Switch.js
```diff
@@ -1,59 +1,64 @@
...
+    if (!navigationState) console.error(`Cannot find a scene with key “${selectedKey}”`);
+
+    if (navigationState.key !== navState.children[navState.index].key) {
+      if (props.unmountScenes) {
+        Actions[selectedKey]({ unmountScenes: true });
       }
-    });
-    if (index === -1) console.error(`A scene for key “${selectedKey}” does not exist.`);
+      Actions[selectedKey]();
```
zxcpoiu: if we have `props.unmountScenes = true`, is it necessary to fire `Actions[selectedKey]()` twice? ( one with `{ unmountScenes: true }` and the other doesn't )

zxcpoiu:
Hmm...I guess it's a redux triggered rendering issue.
I can't use older version of Switch before and never dig into it.
However, I noticed that there is a newer version of Switch recently, and it works.

aksonov:
@zxcpoiu Does @sarovin version works for you?

zxcpoiu:
yeah, it works!

zxcpoiu:
except double render issue, turns out that it can NOT be used with `type="replace"` in the test case above.
thanks guys

zxcpoiu:
**just side note:**
not sure why, the render() times of my component are 2~3 times than the Switch version in 3.32.0
but no problems so far.

sarovin:
@zxcpoiu i try to fix it

sarovin:
@zxcpoiu Can you try now?
@aksonov this is completed for me...

zxcpoiu:
@sarovin

Oops, I can try it 12 hours later.
It looks fine btw.

Thanks for your work!!

aksonov:
"just side note:
not sure why, the render() times of my component are 2~3 times than the Switch version in 3.32.0
but no problems so far." - actually it was initial reason to migrate to new Switch...

blackxored:
This looks promising. Have you tried having a view within a tab trigger an action for one of the external views (the children of the root modal scene)?

sarovin:
@blackxored with modal scene there is a problem.
Can you help me?

blackxored:
I'm not sure it's related to these changes, as I've experienced this before.

sarovin:
@blackxored Yes, it not work with the old version.
So if we have this scenarios:

```
<Scene key="modal" component={Modal}>
      <Scene
        key="root"
        hideNavBar
        component={connect(state => ({ user: state.user }))(Switch)}
        selector={({ user }) => selectorAuth(user)}
        unmountScenes
        tabs
      >
        <Scene
          key="auth"
          rightTitle={'SKIP'}
          rightButtonTextStyle={[Fonts.button, styles.rightButtonTextStyle]}
          onRight={() => { props.dispatch(signInGuest()); }}
        >
          <Scene
            key="welcome"
            component={Welcome}
            title="Welcome"
            initial
          />
          <Scene
            key="signin"
            component={SignIn}
            title="Accedi"
          />
          <Scene
            key="recoveryPass"
            component={RecoveryPass}
            title="Recupera Password"
          />
          <Scene
            key="signup"
            component={SignUp}
            title="Registrazione"
          />
          <Scene
            key="otpScreen"
            component={OTPScreen}
            title="Verifica il numero"
          />
          <Scene
            key="completeProfile"
            component={CompleteProfile}
            title="Completa Profilo"
          />
        </Scene>
        <Scene
          key="main"
          tabs
        >
          <Scene
            key="list"
            component={List}
            title={'Liste'}
            icon={BottomTabIcon}
          />
          <Scene
            key="localize"
            title={'Localizza'}
            icon={BottomTabIcon}
            tabs
            initial
          >
            <Scene
              key="scene1"
              component={Scene1}
              title={'Scene1'}
              icon={TopTabIcon}
            >
              <Scene
                key="search"
                component={Search}
                title="Cerca"
              />
            </Scene>
            <Scene
              key="scene2"
              component={Scene2}
              title={'Scene2'}
              icon={TopTabIcon}
            />
          </Scene>
          <Scene
            key="notifications"
            component={Notifications}
            title={'Notifiche'}
            icon={BottomTabIcon}
          />
          <Scene
            key="profile"
            component={Profile}
            title={'Profilo'}
            icon={BottomTabIcon}
          />
        </Scene>
      </Scene>
      <Scene
        key="error"
        component={Error}
        title="Error"
      />
    </Scene>
```

We can't call Actions.search from Scene with key => Scene1.
Any suggestion?

blackxored:
Do you get scene key errors like I did?

sarovin:
Yes @blackxored

blackxored:
So I wasn't able to fix this. But while I was looking into it, the closest I got was the Switcher setting up a navigation state that only included the child scenes, like assuming it was the root. At the time I thought it was by design and switcher *should be* your root scene, but the tabs requirement was a no-go for me. I'm sure that doesn't help a lot, but hopefully it's a start.

zxcpoiu:
I've tested and it did improved a bit. :+1: 

I'm thinking if it is better we have two Switch implementations, and import it via different use case?
and we can make the most suitable Switch as the default one.

zxcpoiu:
about modal, I don't have nested tabs but FYI:

I can use both `Actions.myModal()`/`Actions.myAnotherScene()`  
in both `tab1 scene` / `tab2 scene` in this Switch implementation

```
    <Scene key="root">
        <Scene key="myModal">
        <Scene key="myAnotherScene">
        <Scene key="switch" initial={true} component={Switch} tabs={true} unmountScenes={true} selector={switchSelector}>
            <Scene key="login" initial={true} />
            <Scene key="home" tabs={true} >
                <Scene key="tab1" initial={true} />
                <Scene key="tab2" />
            </Scene>
        </Scene>
    </Scene>
</Scene>
```

aksonov:
@sarovin Unfortunately it doesn't work for me - infinitive loop (because Actions[selectedKey] re-renders Switch and call updateState again and so so). I'm using mobx and my selector is reactive function. With current version of Switch it works well (because it doesn't call Actions)....

sarovin:
@aksonov I try to fix it

aksonov:
@sarovin Could you just modify current Switch to add Actions call there (if index!=old index) ?

sarovin:
@aksonov 
From this 

```javascript
if (navigationState.key !== navState.children[navState.index].key) {
      if (props.unmountScenes) {
        Actions[selectedKey]({ unmountScenes: true });
      } else {
        Actions[selectedKey]();
      }
    }
```

to this??

```javascript
if (navigationState.key !== navState.children[navState.index].key && navigationState.index !== navState.children[navState.index].index) {
      if (props.unmountScenes) {
        Actions[selectedKey]({ unmountScenes: true });
      } else {
        Actions[selectedKey]();
      }
    }
```

sarovin:
@aksonov You can try with the fix in the last commit?

aksonov:
@sarovin Most interesting thing that  Actions[selectedKey] is never called for my app. I don't understand why selectedKey is always equal to current sceneKey...

aksonov:
When I'm trying just to modify current Switch i see the same - index is always equal to current index.

aksonov:
Could you try following:
```
import React, { Component, PropTypes } from 'react';
import DefaultRenderer from './DefaultRenderer';
import TabBar from './TabBar';
import Actions from './Actions';

export default function Switch(props) {
  const navState = props.navigationState;
  
  const selector = props.selector;
  if (!selector) console.error('Selector should be defined.');
  
  const selectedKey = selector(props);
  if (!selectedKey) console.error('Selector should return key.');
  
  const selected = navState.children.filter(el => el.sceneKey === selectedKey);
  if (!selected) console.error(`A scene for key “${selectedKey}” does not exist.`);
  
  let index = -1;
  navState.children.forEach((el, i) => {
    if (el.sceneKey === selectedKey) {
      index = i;
    }
  });
  if (index === -1) console.error(`A scene for key “${selectedKey}” does not exist.`);
  
  let navigationState = index !== navState.index ? { ...navState, index } : navState;
  if (index !== navState.index) {
    if (props.unmountScenes) {
      Actions[selectedKey]({ unmountScenes: true });
    } else {
      Actions[selectedKey]();
    }
  }
  return (
    <TabBar
      onNavigate={props.onNavigate}
      navigationState={navigationState}
    />
  );
}

Switch.propTypes = {
  navigationState: PropTypes.object,
  onNavigate: PropTypes.func,
  selector: PropTypes.func,
  statem: PropTypes.any,
  unmountScenes: PropTypes.bool,
};
```

sarovin:
@aksonov 
So we have this error:

>  Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.

aksonov:
Maybe Actions should be wrapped into setTimeout also (to avoid warning about setState within render method)

sarovin:
yes, but now the scene is render twice again.

aksonov:
Any idea how to avoid that? I like this variant because it works with my app just fine.

aksonov:
Oh, maybe you could return null if index !== navState.index ?

src/Switch.js
```diff
@@ -1,71 +1,51 @@
...
+    if (props.unmountScenes) {
+      setTimeout(() => {
         Actions[selectedKey]({ unmountScenes: true });
-      } else {
+      }, 10);
+    } else {
+      setTimeout(() => {
         Actions[selectedKey]();
-      }
+      }, 10);
```
aksonov: Ok, will setTimeout with 10 work?

src/Switch.js
```diff
@@ -1,71 +1,51 @@
...
-      return (
-        <DefaultRenderer
-          onNavigate={this.props.onNavigate}
-          navigationState={this.state.navigationState}
-        />
-      );
-    }
-    return null;
   }
+  return (
```
aksonov: Have you tried to return null if selectedKey != current scene? This way no double rendering should happen.

sarovin:
@aksonov I have a problem with the last implementation. Now not work the resetHistoryStack...

sarovin:
I think that we must investigate about your implementation...
I don't understand why selectedKey is always equal to current sceneKey in your use case.

zxcpoiu:
How about pass sceneKey as a prop into TabBar,  and execute it in component Will/Did Mount or will receiveProps if that props exist?

like:
```jsx

    <TabBar
      onNavigate={props.onNavigate}
      navigationState={navigationState}
      fireAction={ () => Actions[selectedKey]() }
    />

```

aksonov:
@zxcpoiu The problem we don't want to render TabBar with old selectedKey.

sarovin:
@aksonov 

Have you any idea about why your scenekey is always egual?
Can you try this?

```
import React, { Component, PropTypes } from 'react';
import DefaultRenderer from './DefaultRenderer';
import Actions from './Actions';

export default class Switch extends Component {

  static propTypes = {
    navigationState: PropTypes.object,
    onNavigate: PropTypes.func,
    selector: PropTypes.func,
    unmountScenes: PropTypes.bool,
  };

  constructor(props, context) {
    super(props, context);
    this.state = {};
    this.updateState = this.updateState.bind(this);
  }

  componentDidMount() {
    this.updateState(this.props);
  }

  componentWillReceiveProps(props) {
    this.updateState(props);
  }

  shouldComponentUpdate(nextProps, nextState) {
    if (this.state.navigationState !== nextState.navigationState) {
      return true;
    }
    return false;
  }

  updateState(props) {
    const navState = props.navigationState;
    const selector = props.selector;
    if (!selector) console.error('Selector should be defined.');

    const selectedKey = selector(props);
    if (!selectedKey) console.error('Selector should return key.');

    const selected = navState.children.filter(el => el.sceneKey === selectedKey);
    if (!selected) console.error(`A scene for key “${selectedKey}” does not exist.`);

    const navigationState = selected[0];
    if (!navigationState) console.error(`Cannot find a scene with key “${selectedKey}”`);

    console.log(navigationState.name, navigationState.sceneKey);
    console.log(navState.children[navState.index].name, navState.children[navState.index].sceneKey);

    if (navigationState.sceneKey !== navState.children[navState.index].sceneKey) {
      if (props.unmountScenes) {
        Actions[selectedKey]({ unmountScenes: true });
      } else {
        Actions[selectedKey]();
      }
    }

    this.setState({ navigationState });
  }

  render() {
    if (this.state.navigationState) {
      return (
        <DefaultRenderer
          onNavigate={this.props.onNavigate}
          navigationState={this.state.navigationState}
        />
      );
    }
    return null;
  }
}

```

aksonov:
No, it doesn't work. The problem that mobx doesn't call componentWillReceiveProps, but just re-render component once observable state is changed. So it works better with pure functions.

Looks like i have equal index because i didn't create deep copy of navigationState...

aksonov:
@sarovin Sorry, last variant works well. So could I merge it ?

sarovin:
@aksonov the variant in the last commit? No, not work the resetHistoryStack. 
Give me time...

sarovin:
@aksonov If this is good for everyone you can merge.

PR #1013, opened 2016-08-04
Restore statem

aksonov:
Restore statem logic used with previous Switch


aksonov:
@sarovin could you please check if it still works for you :)

sarovin:
LGTM

Issue #1014, opened 2016-08-04
How can we pass props to previous screen in Action.Pop() operation


vykumar:
### Version

Tell us which versions you are using: 
- react-native-router-flux v3.30.2
- react-native v0.29.0
### Expected behaviour

I am passing props to previous screen like Actions.Pop({variable: value}) on click of back button. I am not able to get the props in previous screen because there is no any component life cycle methods triggered. 

can anyone please tell me which method is triggered on Action.Pop() operation and what is the way to pass props to previous screen on back button press?
### Actual behaviour

In actual only popping is happened and we moved to previous screen.
### Steps to reproduce

1.
2.
3.

rturk:
@vykumar pop(), by definition, will bring back the previous scene with all the context that it was previously set. 
Why can't you just Go to Actions.MyDesiredScene({withMyNewProps...}); .?

waqas19921:
You can pass props like this 
`NavigationActions.pop( {refresh: {variable: value} })`
this prop change will be found in 
`componentWillReceiveProps `callback in previous component

afilp:
@waqas19921 This is nice, but how can the "automatic" pop of the hardware back (on Android) is handled? What should we do in this case, in order for a prop to have a different value there?

Thanks!

adarshTS:
@waqas19921 how can we get the props in CDM

Issue #1008, opened 2016-08-03
How can I pass variables to previouse scene in Actions.pop() operation


gerard0315:
Hi, I'm using newest router-flux and RN0.30, trying to pass some variables back to a scene's previous scene, i tried put variables like `Actions.pop({variable: this.state.something})` or
`Actions.sceneName({variable: this.state.something, type: ActionConst.BACK_ACTION})` but both failed, can anyone give some suggestions please? thank you

gerard0315:
@doomsower  thx

PR #1016, opened 2016-08-04
Allow POP and similar actions to refresh the previous scene

doomsower:
I recently saw several questions on stackoverflow and here in issues (#1014, #1008), people were asking how to pass props to previous scene on pop. 
So I think it's a good idea to be able to make calls like `Actions.pop({refresh: propsToPassToPreviousScene})`

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,9 @@ import {Actions} from 'react-native-router-flux'
 And then:
 
 * `Actions.ACTION_NAME(PARAMS)` will call the appropriate action and params will be passed to the scene.
-* `Actions.pop()` will pop the current screen. It can also take a param `{popNum: [number]}` that allows to pop multiple screens at once.
+* `Actions.pop()` will pop the current screen. It accepts following optional params:
+  * `{popNum: [number]}` allows to pop multiple screens at once 
+  * `{refresh: {...propsToSetOnPreviousScene}}` allows to refresh the props of the scene that it pops back to
 * `Actions.refresh(PARAMS)` will update the properties of the current screen.
 
 ## Production Apps using react-native-router-flux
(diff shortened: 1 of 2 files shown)


vykumar:
@doomsower , I used the code like Actions.pop({refresh:"farha"}) as you suggested but I am not able to get the props in previous screen because there is no any component lifecycle methods triggered. Now, my question is that like which function/method is triggered when previous screen launch on Action.pop().

doomsower:
@vykumar Are you sure you're testing with my fork of RNRF? If so, can you show some code? Because in my test project everything works fine and componentWillReceiveProps is triggered (you can find this test example [here](https://github.com/doomsower/rn-playground/tree/master/src/rnrf) )

vykumar:
No, I am not using your fork. I just cloned the code from master branch and using version 3.30.2 of RNRF. I used the same code Actions.pop({refresh: {message: this.state.message}}); but it was not working in my project. it's not triggered componentWillRecieveProps.


Can you please tell me what will be the right solution for my project?

doomsower:
Until this PR gets merged (if at all) you have to use my fork, or you can apply this patch to whatever fork you use.

aksonov:
@doomsower Looks nice addition, could you add docs?

doomsower:
@aksonov I've added some info to main readme, near `popNum` param description. But it only documents `pop` action, and not `popTo` and `pushOrPop`, as I feel that main readme should be brief. I think as list of possible action and their parameters grows, a section should be created which describes all this stuff.

teyou:
so this PR will get bumped into version "3.33.3" ?
Looking forward to try this :)

juanpasolano:
Hi, Any idea when is this going to be publish to npm? 
Thank you

aksonov:
done

conor909:
This seems to be an issue again: [#1381](https://github.com/aksonov/react-native-router-flux/issues/1381)

PR #1021, opened 2016-08-04
ADD react-addons-pure-render-mixin dependency

sarovin:


aksonov:
Any difference between 15.2.1 and 15.0.2?

sarovin:
react-addons-pure-render-mixin@15.2.1 requires a peer of react@^15.2.1

PR #1032, opened 2016-08-08
Add 'leftToRight' direction support

stephy:
This will allow users to choose 'leftToRight' direction.

Ex:
```javascript
<Scene key="settings" component={Settings} direction="leftToRight" title="Settings"/>
```

### DIFF
--- a/src/DefaultRenderer.js
+++ b/src/DefaultRenderer.js
@@ -14,6 +14,7 @@ import {
   Animated,
   View,
   StyleSheet,
+  Dimensions
 } from 'react-native';
 
 import TabBar from './TabBar';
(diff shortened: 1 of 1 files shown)


aksonov:
LGTM

aksonov:
@stephy Could you please add new mode to docs?

wrousseau:
@stephy Hi ! What is the difference with the standard direction used by default ?

stephy:
@wrousseau correct me if I am wrong, but the default direction is from right to left. For my app I really wanted from left to right. Since I added this feature to my app, I figure I would just do a PR as other people may want it from left to right as well. Is there a way to do navigate from left to right? I could not find it.

stephy:
@aksonov I will do that now.

wrousseau:
@stephy Alright, got it. I feel like the name could be confusing though because locale whose transitions between scenes are supposed to be following that direction are usually called "right to left" locales. There is actually some work being done on React Native to support rtl languages

stephy:
@wrousseau I see what you mean, I guess that can really get confusing. I could change the name to 'horizontalLeftToRight'. What do you think?

joenoon:
I haven't been able to try this PR yet so I might not understand exactly what we're discussing (i.e. feel free to ignore me if this doesn't make sense).  But if it does what I'm guessing, then what about: `appearFromLeftToRight`

stephy:
@joenoon I think that's a great suggestion

PR #1043, opened 2016-08-10
added getRightTitle and getLeftTitle

wtfil:
Hi!

I faced with problem that left and right buttons could not have dynamic values based on app state. 
I added two more optional properties `getRightTitle` and `getLeftTitle` which works in the same way as `getTitle`


RainThinks:
exactly what I am looking for. hope it can be merged soon. 
my kudos to @wtfil

PR #1049, opened 2016-08-12
Feat(BackAndroid): Add handler for Back Key in Android

Swordsman-Inaction:
# Propose a better solution for handling android back key.
I proposed a solution before: #820 , but not good.

### Now I add a default handler for android back key:
It will pop a scene after back key is pressed, and will exit the app when there is no scene to pop.
I think it's good for most app.
### And it provides some functions to customise the handler:
**backAndroidHandler**: you can pass a function to override the default handler, do whatever you want.
**onBackAndroid**: get noticed on every pop action, but not affect the default behavior.
**onExitApp**: pass a function to change the behavior when there is no scene to pop. Return `true` to stay in app or return `false` to exit the app. Here you can do something like pressing twice to exit app.

### Need your opinion:
Should this default handler be enabled by default or disabled by default? @aksonov @joenoon


joenoon:
@Swordsman-Inaction looks pretty good to me.  I haven't had a chance to run it... have you tried to run on iOS with these new Router additions? I'm curious if BackAction will complain since its an Android lib, or if its just a no-op.

Swordsman-Inaction:
@joenoon works good on my project. `handleBackAndroid` won't be called on iOS, since `BackAndroid.addEventListener('hardwareBackPress', this.handleBackAndroid)` just works on android.

Kerumen:
@aksonov Can you release a npm version with this? Thanks!

veedeo:
waiting for npm release aswell and thanks for the feature

Issue #994, opened 2016-07-31
renderRightButton displays button on left side of NavBar


AlessandroCorradini:
### Version

Tell us which versions you are using: 
- react-native-router-flux v3.32.0
- react-native v0.30.0
### Expected behaviour

I would to render a custom icon for the right button with renderRightButton method. I'm using react-native-vector-icons. The button appears on left. The button works, but it is of the wrong side of the NavBar.
### Actual behaviour

![screenshot_20160731-164100](https://cloud.githubusercontent.com/assets/13152962/17277327/9ca9950e-5740-11e6-90c7-c7dfe7fa9b73.png)
### Steps to reproduce

`const MenuIcon = <TouchableHighlight onPress={()=>Actions.menu()} style={{ backgroundColor: 'red', width: 30, height: 30}}>
                    <Icon name="menu" size={30} />
                 </TouchableHighlight>;`

`<Scene key="home" component={HomePage} title="Home"  onRight={()=>Actions.menu()} rightTitle="Menu" initial={true} renderRightButton={() => MenuIcon}/>`

zth:
I have the same issue :)

waqas19921:
I have the same issue :)
if i move it to right by giving `paddingLeft` the touch listener on left button also trigger right button function.

sarovin:
Can you try to fix it with style props?

AlessandroCorradini:
@sarovin 
yes, if I add "position: absolute" and "right: 0" or another number, the button is in the correct position, but I think than the **right** button should be on the right side by default and the style props should serve for micro adjustments :) 

```
<TouchableHighlight onPress={()=>Actions.menu()} style={{ backgroundColor: 'red', width: 40, height: 40, right: 5, top: 5, position: 'absolute'}}>
        <Icon name="menu" size={40} />
    </TouchableHighlight>;
```

![screenshot_20160804-220954](https://cloud.githubusercontent.com/assets/13152962/17416999/4474f550-5a91-11e6-846c-b7ebb520167d.png)

moschan:
I sent Pull request (https://github.com/aksonov/react-native-router-flux/pull/1054) for solve this problem, and It was merged. Please check this function when next version bumped :smile_cat:

marciobueno1:
My navigationBarStyle is
`{
    flex: 1,
    flexDirection: 'row',
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: 'transparent',
[...]

PR #1054, opened 2016-08-14
Add default style to renderLeftButton and renderRightButton

moschan:
## Abstract
Added default style to `renderLeftButton` and `renderRightButton` for reduce extra adjustments.

## Detail
`renderLeftButton` and `renderRightButton` provide a method to add a component to navigation-bar. But, If we add component without adjusting style, It will be shown unexpected position. Example is below:

screenshots_2016-08-14_17_08_25

So, I added view component that has default style as parent component of these components. Result is below:

screenshots 2016-08-14 17 05 52

## Related Issue
- #994

### DIFF
--- a/Example/Example.js
+++ b/Example/Example.js
@@ -28,19 +28,6 @@ import EchoView from './components/EchoView';
 import NavigationDrawer from './components/NavigationDrawer';
 import Button from 'react-native-button';
 
-const Right = () => (
-  <Text
-    style={{
-      width: 80,
-      height: 37,
-      position: 'absolute',
-      bottom: 4,
-      right: 2,
-      padding: 8,
-    }}
-  >Right</Text>
-);
-
 const styles = StyleSheet.create({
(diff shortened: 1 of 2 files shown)


aksonov:
@joenoon, @sarovin Does this change work well with existing renderButton functions in your apps?

joenoon:
I have a custom nav so I won't be affected.  My guess is people using the default nav will need to make adjustments based on this.

aksonov:
@moschan Ok, i will apply for now, but please keep eye on possible issues for users with defined renderButton functions.

moschan:
@aksonov Thank you for your review! OK, I will check issues carefully.

marciobueno1:
My navigationBarStyle is 
{
    flex: 1,
    flexDirection: 'row',
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: 'transparent',
    borderBottomWidth: 0,
  }
So my renderRightButton style is broken when I use 3.36.0, it is fine with 3.35.0

PR #1057, opened 2016-08-15
Support passing props to tab routes

benstepp:
Basically when a JUMP occurs, a copy of the wrapped scene is pulled out of state.scenes and placed in the state tree with the props copied onto it. When no props are passed in the action, all props are removed as every jump grabs a new copy of the scene. 

I'm not too sure about the implementation, so any other ideas may be worth exploring.

```jsx
<Scene key="root" tabs>
  <Scene key="home" component={Home} />
  {/*...*/}
</Scene>

Actions.home({customProp: true})
// customProp = true
// DefaultRenderer renders Home with customProp: true

Actions.home()
// customProp = undefined
// DefaultRenderer renders Home without customProp key in props
```

Reducer.test.js has the tests relevant to this change. I was just familiarizing myself with the codebase in Actions.test.js.

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -76,6 +76,7 @@ function inject(state, action, props, scenes) {
     return state;
   }
   let ind;
+  let children;
 
   switch (ActionMap[action.type]) {
     case ActionConst.POP_TO: {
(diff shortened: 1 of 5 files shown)


src/Reducer.js
```diff
@@ -163,7 +163,18 @@ function inject(state, action, props, scenes) {
...
+
+          // Pass the props to the wrapped tab scene
+          if (props.children) delete props.children;
+          const key = child.children[0].sceneKey;
+          const wrappedScene = scenes[key];
+          child.children = [{ ...props, ...wrappedScene }];
+        }
+      });
+
       assert(ind !== -1, `Cannot find route with key=${action.key} for parent=${state.key}`);
```
aksonov: You should not modify existing state, but produce new one. Here i see you modify existing child within existing state var.

benstepp: Good looking out. It just seemed like the way to go with all of the shallow slices pointing back to previous state objects, and direct setting of state vars [here](https://github.com/aksonov/react-native-router-flux/blob/master/src/Reducer.js#L178) and [here](https://github.com/aksonov/react-native-router-flux/blob/master/src/Reducer.js#L191).

This does give me a change to better cover edge cases though. I noticed when nesting tabs things stop behaving as expected. I'm thinking at least tests for double nested tab routes should cover it. I'll take another stab at it this weekend.

nicholeuf:
This feature would be very nice to have!

Herlevsen:
This is not working anymore

PR #1059, opened 2016-08-15
pass direction to getPanHandlers

vonovak:
pass direction to custom getPanHandlers function


philipshurpik:
@vonovak good idea to pass direction, but please fix eslint :)

aksonov:
@charpeni Any idea why PRs fail unit tests now?

Issue #1092, opened 2016-08-23
Back button is not shown when tabs scene is pushed


fkoester:
I have a scenario where I need tab views nested inside another tab view. When pushing the child tabs view, the scene is correctly pushed but the navbar does not show the back button. On Android I can correctly go back using the back button, though.  
### Version

Tell us which versions you are using: 
- react-native-router-flux v3.34.0
- react-native v0.31.0
### Expected behaviour

The back button should be shown when I can go back.
### Actual behaviour

The back button is not shown when pushing a tabs scene.
### Steps to reproduce

I created a little project demonstrating the problem, the main file is [here](https://github.com/fkoester/react-native-router-flux-nested-tabs/blob/8a7acc40fd9d5c6e79b44e5ff2a00f3a9d38030d/NestedTabsDemo.js).
1. Navigate to the tabbar view
2. On the tab, select to push the next tabs view
3. The push is performed, but the back button is not shown. You can go back by Actions.pop(), though (like on Android using the back button).

fkoester:
Please see my proposed fix in PR https://github.com/aksonov/react-native-router-flux/pull/1093.

PR #1093, opened 2016-08-23
Fix for back button not displayed when pushing tabs scene.

fkoester:
This is a proposed fix for bug https://github.com/aksonov/react-native-router-flux/issues/1092.

When pushing a tabs scene, we add another prop `parentIndex` to the child scene. The navbar can then check the value of it in order to decide, whether to show the back button or not.

### DIFF
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -206,7 +206,7 @@ class NavBar extends React.Component {
       childState.leftButtonStyle,
     ];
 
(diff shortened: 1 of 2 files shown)


charpeni:
Can you fix eslint?

fkoester:
@charpeni Sure! Sorry, is fixed now. I also rebased against latest master.

PR #1094, opened 2016-08-23
Allow to force parent tabs to hide when child tabs is added

fkoester:
I have a scenario where I need to go from one tab to another `tabs` scene using push (so the user can also  go back).

If you have a scenario like that, I can imagine two desired behaviors:

1. Both tabbars are displayed among each other (Is quite uncommon though). This way you can always switch tabs on both levels.
2. As soon as the child tabbar is displayed, the parent tabbar is hidden so you always have only one tabbar.

My proposed solution is therefore to introduce another flag I called `hideOnChildTabs`, which - when set to `true` on the parent `tabs` scene -  causes the second behavior. Otherwise the first behavior is used and you have to manually take care of setting your stylesheets so that they are displayed properly on the same screen.

Without this patch, I experienced that the parent tabbar always covered the children's tabbar, which can never be the intended behavior.

I created a little project demonstrating the situation [here](https://github.com/fkoester/react-native-router-flux-nested-tabs).


charpeni:
Can you fix eslint?

fkoester:
@charpeni Fixed issue reported by eslint and rebased.

aksonov:
Could you please solve conflicts? Thanks

fkoester:
@aksonov Sure, done ;-)

charpeni:
@aksonov CI is fixed on master, you can safely merge this PR if you wish to.

aksonov:
Thanks, but looks like master has some weird issues with Switch and some other, could you check?

Pavel.

fkoester:
I rebased against the latest master and added an entry to the API documentation.

Issue #376, opened 2016-03-23
Key login is already defined - HMR 0.22


PeterKow:
I am on `react-native 0.22` and `react-native-router-flux 3.1.3` with hot reloding [HMR], after first HMR reload I got following error 

![image](https://cloud.githubusercontent.com/assets/10186048/13956908/730928a0-f042-11e5-8c2a-988066f2b0f9.png)

Problem is caused by `Actions.create`

This is my current setup:
```javascript
let scenes = Actions.create(
     <Scene key="root">
            <Scene key="login" component={Login} title="Login"/>
          </Scene>
);

 <Provider store={store}>
      {() =>
        <Router scenes={scenes} />
      }
    </Provider>
```

I tired this also, but doesn't work at all - nothing renders
```javascript
 <Provider store={store}>
      {() =>
        <Router>
          <Scene key="root">
            <Scene key="login" component={Login} title="Login"/>
          </Scene>
        </Router>
      }
    </Provider>
```

Temporary solution is to call when working with HMR is to call components directly without Router

```javascript
{() =>
        <Login/>
      }
..

Let me know if you have any ideas, so I can test them :)

**Comment by aksonov** [comment]
Check latest version

**Comment by zxcpoiu** [comment]
same issue using 3.1.5

**Comment by aksonov** [comment]
It is not true. 3.1.5 doesn't throw exception for existing keys anymore
```
 if (this[key]){
            console.log("Key " + root.key + " is already defined!");
        }
```

zxcpoiu:
oh sorry, I mean, I didn't got a red screen but same issue which is "key xxx is already defined."

aksonov:
It is not issue but logging. It still could be usable if user really defines several scenes with the same key. For hot reload you could ignore it.

zxcpoiu:
@PeterKow

I think it is caused by wrap `<Router>` into `<Provider>` and maybe it is redux triggers `<Router>` to re-render.
you can try Example and print some log in `render()` and `onComponentDidMount`, you will find that `<Router>` only render at the initial stage.

is your `<Router>` render several times also?
I'm still trying to find a better way to wrap `<Provider>`

PeterKow:
Works after updating to 3.1.5

@zxcpoiu I tried without Provider and that wasn't a problem

Thanks!

agrass:
I'm still having this issue. Any idea how to solve it? I have the camera in one scene and is rendering multiples times and finally the app crash due low memory

[...]

PR #1098, opened 2016-08-24
update note to prevent `Key is already defined`

zxcpoiu:
for redux users who have architecture problems integrating redux and router flux.


zxcpoiu:
#376

PR #1099, opened 2016-08-24
fix Example Android marginTop

xiaokekeT:
Is this correct?


MechanicKim:
I think it is correct.
I tested it on Android and iOS and I confirmed that it is correct.

PR #1119, opened 2016-08-30
Add action const POP_AND_REPLACE

bifrost:
I needed a way to change the scene stack `[a, b, c] -> [a, d]` so swipe BACK action natural went from scene `d` to `a`. It is not possible with existing `actionConst` in one go. Therefore I had to create the new `actionConst` POP_AND_REPLACE that combine BACK action with REPLACE action, where `popNum` is optional and default 1.


linonetwo:
Can you release it on npm? Badly need it!

Naoto-Ida:
+1 would be a great addition for what I'm developing right now :)

PR #1129, opened 2016-09-01
Revert "Support passing props to tab routes"

aksonov:
Reverts aksonov/react-native-router-flux#1057


rseemann:
@aksonov why was this removed?

PR #1132, opened 2016-09-02
export TabbedView

n7olkachev:
Useful when you want to create custom tabs component


kachizukanne:
Thanks @charpeni 

When does the npm version get this update?

PR #1133, opened 2016-09-02
Revert "Revert "Support passing props to tab routes""

aksonov:
Reverts aksonov/react-native-router-flux#1129

### DIFF
--- a/src/Actions.js
+++ b/src/Actions.js
@@ -203,4 +203,5 @@ class Actions {
   }
 }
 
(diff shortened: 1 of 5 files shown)


aksonov:
Unit test fails, I thought because of this PR, but it is not the reason... I will merge it again.

jfo84:
Would love a new versioning for this, @aksonov. It's not optimal to pull it from Github 😄 Thanks in advance!

Issue #686, opened 2016-05-16
Actions.callback is not defined!


dominicmh:
### Version
- react-native-router-flux v3.22.8
- react-native v0.24.0-rc5

### Issue
I'd expect that as soon as the Router is rendered, Actions has a callback for all scenes to transition to.

In my app I tried to transition to a login scene if the user isn't logged in at start. However, I get an error as soon as Actions.login() in the componentDidMount() function is executed: 
> Actions.callback is not defined!

If I try to do the same from within the Main.componentDidMount() function it works as expected.

At which point are the Scene callbacks added to Actions? Am I doing it wrong?

```
const reducerCreate = params => {
  const defaultReducer = Reducer(params);
  return (state, action) => {
    console.log('ACTION:', action);
    return defaultReducer(state, action);
  };
};
componentDidMount() {
  if (!loggedIn()) {
    Actions.login();
  }
},
render() {
  return (
    <Router createReducer={reducerCreate}>
      <Scene key="root" hideNavBar>
        <Scene key="login" component={Login} />
        <Scene key="main" component={Main} initial />
      </Scene>
    </Router>
  );
}
```

rturk:
@dominicmh 
FYI https://github.com/aksonov/react-native-router-flux/issues/525

Considering https://github.com/aksonov/react-native-router-flux/issues/636 and that RN26 will be released within the next couple of days please close this issue

charpeni:
I'll close this issue as @rturk said 0.26.0 version will be landed soon. (react-native and react-native-router-flux)

If you still get this issue with 0.26.0, please reopen it.

dominicmh:
I updated to RN 0.26.0, but the issue remains:
> [react-native-router-flux] Actions.callback is not defined!

I used the 0.26-wip branch as suggested here [#636](https://github.com/aksonov/react-native-router-flux/issues/636#issuecomment-220253826) in order to make RNRF runnable on RN 0.26.0

* "react": "^15.0.2",
* "react-native": "^0.26.0",
* "react-native-router-flux": "github:aksonov/react-native-router-flux#0.26-wip"

rturk:
@dominicmh paste more details of your code in a Gist.

dominicmh:
[Fails in App.js, note that it works fine in MissionsTab.js](https://gist.github.com/dominicmh/f86be050221c0568d1360a40c4c04301)

rturk:
@dominicmh My Two cents
[...]

PR #1137, opened 2016-09-03
fix routes not initialized on parent componentDidMount

atlanteh:
Since setState is async then on componentDidMount of parent the router is not initialized yet and running Actions.someRouter() results in Actions.callback is not defined exception.

### DIFF
--- a/Example/__tests__/__snapshots__/Example-test.js.snap
+++ b/Example/__tests__/__snapshots__/Example-test.js.snap
@@ -174,9 +174,9 @@ exports[`Example renders correctly 1`] = `
                       Launch page
                     </Text>
                     <View
-                      accessibilityComponentType={undefined}
+                      accessibilityComponentType="button"
                       accessibilityLabel={undefined}
-                      accessibilityTraits={undefined}
+                      accessibilityTraits="button"
                       accessible={true}
                       hitSlop={undefined}
                       onLayout={undefined}
@@ -211,9 +211,9 @@ exports[`Example renders correctly 1`] = `
                       </Text>
                     </View>
                     <View
-                      accessibilityComponentType={undefined}
+                      accessibilityComponentType="button"
                       accessibilityLabel={undefined}
-                      accessibilityTraits={undefined}
+                      accessibilityTraits="button"
                       accessible={true}
                       hitSlop={undefined}
                       onLayout={undefined}
@@ -248,9 +248,9 @@ exports[`Example renders correctly 1`] = `
                       </Text>
                     </View>
                     <View
-                      accessibilityComponentType={undefined}
+                      accessibilityComponentType="button"
                       accessibilityLabel={undefined}
-                      accessibilityTraits={undefined}
+                      accessibilityTraits="button"
                       accessible={true}
                       hitSlop={undefined}
                       onLayout={undefined}
@@ -285,9 +285,9 @@ exports[`Example renders correctly 1`] = `
                       </Text>
                     </View>
                     <View
-                      accessibilityComponentType={undefined}
+                      accessibilityComponentType="button"
                       accessibilityLabel={undefined}
-                      accessibilityTraits={undefined}
+                      accessibilityTraits="button"
                       accessible={true}
                       hitSlop={undefined}
                       onLayout={undefined}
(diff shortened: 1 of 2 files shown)


atlanteh:
#686

atlanteh:
guys any update regarding this?

joenoon:
@atlanteh I've looked at this diff and read through the comments on https://github.com/aksonov/react-native-router-flux/issues/686 

I'm pretty sure this is a matter of moving the code that calls Actions.foo() into the right spot.  It looks like right now at least from the original issue, Actions.foo() is being attempted in a parent and is expecting its children to execute (create the reducer) before it can function, which is against how the react component lifecycle works.

This probably means using a Switch to handle the scene picking in the context of the router and its lifecycle.  Or at the very least moving the Actions.foo() to a component used *by* the router, not in the component that is responsible for creating the router.

atlanteh:
@joenoon
Actually, I ran into that problem because I ran into the same one.Calling an action in rnrf parent's componentDidMount results in an error. We choose the screen dynamicly, so we call different routes on componentDidMount. It's expected that the router is initialized when it's mounted and it's not. Only after the router is mounted it starts to initialize itself *asynchronously*, which means we have no idea when it's safe to call an action.
It makes no sense that we must set arbitrary timeout in hopes that the router is initialized after it, and currently this is the conculsion of #686.
My PR makes sure that once the router is mounted it's also initialized.
This allows the parent to call an actions on componentDidMount.

joenoon:
@atlanteh a couple thoughts:

First, I'm not against the PR as long as people have tested it.  I'm currently way behind versions so I can't test this on my app.  Hopefully @aksonov and others can review.

But some other thoughts: `Actions` tends to get treated like a singleton, yet Router is not.  This has always been an oddity/bug IMO, but most of the time people don't run into issues because they use only one Router.  But really, I don't think there should be a global `Actions`.  If the router exposed a `actions` method and passed that down either through context or props, then the situation here would be more clear.  You would need to access `this.props.actions` which would make it obvious where `actions` is available.  I'm not sure, but I would hope that would take the async/race issues out of this.

If you do want to treat `Actions` as a singleton, then my suggestion would be to define your actions all up front:

```
const scenes = Actions.create(
  <Scene key="root" hideNavBar hideTabBar>
    ... scenes ...
  </Scene>
);

export const MyRouter = function (props) {
  return <Router key="router" scenes={scenes} ... />;
}
```

Again, not sure if that would fix this specific problem, but just throwing ideas out there.

atlanteh:
@joenoon, the problem is not the routes not being initialized, they actually are initialized synchronously. The problem is the router itself.
You see, Actions.callback is only defined in [renderNavigation](https://github.com/aksonov/react-native-router-flux/blob/master/src/Router.js#L128).
The problem is, that renderNavigation is called only after [this.state.reducer](https://github.com/aksonov/react-native-router-flux/blob/master/src/Router.js#L144).
So using Actions as a singleton or not, this problem is not solved.

atlanteh:
Ok, so I just cloned my repository with the change and it seems to work great with your example. I even played a little bit with hot loading and changing names & backgrounds and it doesn't seem to break any functionality. You can try it by cloning:
git clone -b "patch-1" https://git@github.com/atlanteh/react-native-router-flux.git
and then trying your example.
It would help a lot if the guys from #686 can try it and confirm it's working.
@dominicmh, @rturk @charpeni @aksonov @dwilt @shilpan
Can you guys help?

joenoon:
I just pulled this down and tried it in my app.  I couldn't really do much to test it since there is a pretty bad bug in master that seems to be wiping out the tab bar, making master basically broken.  (Anyone know what happened there?)

I have a feeling by looking at the changes that this is ok.  And worst case, it won't be the only blocker on master.  I'll give it a few hours and then merge this unless anyone objects before then.

chetstone:
@joenoon I discussed the tab bar regression in issue #1151 . I identified the offending commit. This is a pretty serious problem that has not been addressed.

aksonov:
Ok, let's revert it

Pavel.

dwilt:
Are you reverting this fix or something else not related to this ticket?

aksonov:
@dwilt Regression made https://github.com/aksonov/react-native-router-flux/issues/1151

dwilt:
@aksonov I'm confused. So are you saying this commit (https://github.com/aksonov/react-native-router-flux/commit/492a1171b8606869e5808f48220e7b0df4699c88) that got merged caused a regression (https://github.com/aksonov/react-native-router-flux/issues/1151) and you're going to revert this merge?

Not arguing, just want to make sure I understand what is going on and if I need to revert this patch in our project.

aksonov:
@dwilt Check that comment ("@joenoon I discussed the tab bar regression in issue #1151 . I identified the offending commit. This is a pretty serious problem that has not been addressed.")

@charpeni, @joenoon Looks like @chetstone is not available, could you look? We can't publish new release because of that bug...

PR #1140, opened 2016-09-04
Fix syntax error in example

RafalWilinski:
Add missing `)` in drawer example.


charpeni:
Nice catch, thank you!

PR #1150, opened 2016-09-07
Export TabbedView for use in custom TabBar

chetstone:
TabbedView is useful when creating a custom TabBar component based on RNRF TabBar


charpeni:
Duplicate of #1132.

PR #1153, opened 2016-09-08
adding snappatizer to readme

erdostom:
We just launched our new app that uses `react-native-router-flux`!


charpeni:
Congratulations!

erdostom:
Thank you! :)

Issue #1167, opened 2016-09-14
Can't install the Example on Android


compojoom:
Hey guys!

I was able to run the example in my own app (just copied the js files and everything seems to work with them). However I'm not able to clone the repository and run the examples in their own app. During the build process I get:

```
Example\android\app\build\outputs\apk\app-debug.apk: Failed to finalize session : -26: Package com.example new target SDK 22 doesn't support runtime permissions but the old target SDK 23 does.
```

Has anyone run into this issue?
I'm trying to install this on a real device running android 6

compojoom:
I couldn't find any app called com,example on my device, but after changing the app package to com.aksonov.routerflux.example I was able to build and deploy to my device.

compojoom:
Closing as my pull request was merged

PR #1168, opened 2016-09-14
fixing issue #1167

compojoom:
It seems that the android package uses a too generic package name. Changed this to have a distinct namespace com.aksonov.routerflux.example


philipshurpik:
Thanks for android example improvements!

PR #1172, opened 2016-09-15
Allow the passing of props via the popTo Action

rseemann:
This simple change allows the `popTo` action, which is now called only with the scene name as parameter, to receive an object that carries the scene name and can be used for refreshing props, just as the other `pop` actions.

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -340,8 +340,15 @@ function reducer({ initialState, scenes }) {
 
       // Find the parent and index of the future state
       if (ActionMap[action.type] === ActionConst.POP_TO) {
-        const target = action.data;
-        assert(target, 'PopTo() must be called with scene name');
+        /*
+         * if a string is passed as only argument
+         * Actions.filterParam will put it in the data property
+         * otherwise look for the scene property
+         */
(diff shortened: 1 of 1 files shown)


brunolemos:
@rseemann How to do that exactly?

I tried this: `Actions.popTo({ scene: 'myPage', myPropToUpdate: [1, 2, 3]});`

Where `myPropToUpdate` is the new value I want the page `myPage` to refresh.

rseemann:
Is the action poping to the right scene and the `myPropToUpdate` is not updating, or what is happening?

brunolemos:
Got it working!
The props must be in the `refresh` params and `componentWillReceiveProps` must be implemented.

`Actions.popTo({ scene: 'myPage', refresh: { myPropToUpdate: [1, 2, 3] }});`

`myPage.js`
```
componentWillReceiveProps({ myPropToUpdate }) {
    if (myPropToUpdate !== this.state. myPropToUpdate) {
      this.setState({ myPropToUpdate });
    }
  }
```

cc @rseemann

rseemann:
@brunolemos oh, glad to hear it!

PR #1190, opened 2016-09-18
code refactorings

stereodenis:


charpeni:
Unfortunately, class properties are not ES6 and even less a standard yet, it's only a proposal at stage 2.

I would not recommend using this until stage 3 at least.

sibelius:
arrow functions looks better than `.bind()`

there are a lot of people using class properties on react native projects, we can easily change this if something goes wrong and it did not reach stage 3

charpeni:
I understand what you mean and you're right, arrow functions looks better than `.bind()`.

However, a lot of people doing something doesn't necessarly justify it. We're talking about a router library and for the sake of stability I'd like to not rely on babel quirks. Issues are growing up since few months and few contributors have time to help, adding this to the pipeline without a proper description, nor even a test plan seems a bit crazy to me.

Using this in a project? Go ahead. 
In a library that many users rely on? Sorry, but that will be a no for me.

@stereodenis Anyway, thank you for your contribution it's highly appreciated.

I'll let @aksonov close or merge this pull request, it's his library after all.

charpeni:
@aksonov ?

stereodenis:
@charpeni @aksonov I had rebased branch

aksonov:
@charpeni I don't understand advantage of this refactoring...

stereodenis:
@aksonov 

- readability
- new modern syntax
- less code

aksonov:
Sorry, but 
`renderLeftButton = (navProps) => {` is not more 'readable' than `renderLeftButton (navProps) {`

The only one reason to do it is to avoid manual binding, but it doesn't make code more readable...

stereodenis:
@aksonov yep, https://daveceddia.com/avoid-bind-when-passing-props/

charpeni:
>**Way #4: Class Instance Field With an Arrow Function (good, ES8+)**
>Even though this is supported by Babel, there’s a (small) risk that this feature could be taken out of the spec and require some refactoring, but a lot of people are using it so it seems likely that it’ll stay put.

>**Way #5: Binding in the Constructor (good, ES6)**

As I said in my [previous message](https://github.com/aksonov/react-native-router-flux/pull/1190#issuecomment-253651964), for the sake of stability I'd like to not rely on babel quirks. (ES8+).

PR #1215, opened 2016-09-24
backgroundImage style

openmynet:


sarovin:
@openmynet  Can you fix lint error?

sibelius:
can't we make backgroundImageStyle as a prop to be customizable?

PR #1219, opened 2016-09-26
change navbar

AkiraXue:
change navbar for which i use component to do the leftButton and aRightButton


sibelius:
every style should be customizable

sarovin:
@aksonov you can close

PR #1232, opened 2016-09-29
Provide routes to scene's context

peteychuk:
This PR makes it so that **Router** passes **routes** as context to the rendered scene.

I believe this to be a valid use case for context. Would be much appreciated if we could get this merged in.

Thanks! Let me know if you have any questions.

### DIFF
--- a/Example/__tests__/__snapshots__/Example-test.js.snap
+++ b/Example/__tests__/__snapshots__/Example-test.js.snap
@@ -197,7 +197,7 @@ exports[`Example renders correctly 1`] = `
                           Array [
                             Object {
                               "color": "#007aff",
-                              "fontFamily": ".HelveticaNeueInterface-MediumP4",
+                              "fontFamily": "HelveticaNeue-Medium",
                               "fontSize": 17,
                               "fontWeight": "bold",
                               "textAlign": "center"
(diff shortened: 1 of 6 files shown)


sibelius:
why not just use `Actions` global?

peteychuk:
It's good to use Actions global.
But in my opinion, it would be a good opportunity to get Actions from context for complex applications. I believe this to be a valid use case for context.
For example:
```
// components/MyComponent.js
import React, { PropTypes, Text, View } from 'react-native';
import { connect } from 'react-redux';
import Button from 'react-native-button'

class MyComponent extends React.Component {
  static contextTypes = {
    routes: PropTypes.object.isRequired,
  }

  render () {
    const {routes} = this.context;

    return (
      <View>
          <Button onPress={()=>routes.login({data:"Custom data", title:'Custom title' })}>Go to Login page</Button>
          <Button onPress={routes.register}>Go to Register page</Button>
          <Button onPress={routes.back}>Go back</Button>
      </View>
    );
  }
}

export default MyComponent;
```

aksonov:
@peteychuk Makes sense, but could you state in update docs that it is optional step? Most users may want to use Actions global directly.

peteychuk:
@aksonov Agree. Documentation has been updated.

PR #1244, opened 2016-10-04
Improved RNRFActions interface definition

jankalfus:
I've improved the RNRFActions interface definition so that the `Actions` object is indexable by key. In code, we can then use `Actions["mySuperAwesomeScene"]()` to make a transition. Better yet, you can create an enum (e.g. `Scenes`) and corresponding `getKey` function which returns scene key based on the enum value, and use it like this: `Actions[getKey(Scenes.Register)]();`

### DIFF
--- a/docs/OTHER_INFO.md
+++ b/docs/OTHER_INFO.md
@@ -22,7 +22,7 @@ import StatusModal from './components/StatusModal'
       <Scene key="screen1" initial={true} component={Screen1} />
       <Scene key="screen2" component={Screen2} />
     </Scene>
-      <Scene key="statusModal" component={StatusModal} />
+    <Scene key="statusModal" component={StatusModal} />
   </Scene>
 </Router>
 ```


sibelius:
looks good

dalcib:
Excelent. I was trying to solve this. Thank you.

jankalfus:
Is there an issue? Can we merge pls?

aksonov:
@jankalfus CircleCI fails tests, could you check it?

jankalfus:
@aksonov There is no reason why they should fail. I've changed two lines in a typescript definition file, nothing else. From the CircleCI report it looks like it ran test for a completely different commit.

wcandillon:
@jankalfus Is it possible to have a new version of the package that publishes this change?

charpeni:
@wcandillon It should be available in the latest release on npm (3.37.0).

wcandillon:
@charpeni thks a lot!

PR #1261, opened 2016-10-10
Fix the style of the NavBar content wrapper and its customizable children

jgkim:
Currently, the content of the NavBar is wrapped by `View` so you cannot style the components in it with relative position. To fix this, the default style for the wrapper has changed, and customized styles are also applied to cutomized components. With this PR, you can use `render*Button` prop with the root scene.


aksonov:
I'm really sorry about missing this PR. Unfortunately I have available time only recently and completed v4 beta based on React Navigation. Feel free to submit PRs for v4 now. I promise I will not miss them again.

PR #1287, opened 2016-10-14
Add more detail information on TabBar

dragfire:
More detail explanation on Customizing TabBar and Refreshing a particular Scene when a Tab is pressed.

Hope this would be helpful to someone.


charpeni:
Thank you!

Issue #1311, opened 2016-10-21
Initial tab will be frozen when change tab


phuongpt:
### Version

Tell us which versions you are using: 
- react-native-router-flux v3.36.0
- react-native v0.32.0
### Actual behaviour

The initial tab will be frozen, can't scroll or do anything when we change to another tab then back.

Kaishley:
#1270

zhanbst:
Use this event at initial tab  Actions.calendar({type: ActionConst.REFRESH});
```
<Scene key="myTabBar" tabs={true} hideNavBar tabBarStyle={style.tabBarStyle}>
  <Scene 
    key="myTab" 
    title="My Tab" 
    icon={MyTabIcon} 
    onPress={()=> {
      Actions.myTab_1({type: ActionConst.REFRESH});
    }}
   >
      <Scene key="myTab_1" component={MyTabComponent} hideNavBar/>
  </Scene>
</Scene>
```

phuongpt:
thanks @zhanbst

WillCoco:
thanks

leonacky:
@zhanbst Not work with latest version

Issue #1293, opened 2016-10-17
Initial Component in TabBar Mounted Twice


dmhood:
### Version
Tell us which versions you are using: 

- react-native-router-flux v3.36.0
- react-native v0.34.1

### Expected behaviour
Navigating between tabbed scenes mounts each component a single time.

### Actual behaviour
The initial component in a group of tabbed scenes is mounted a second time if you navigate to a different tab and navigate back.

### Steps to reproduce

1.  Download and run the current example, replacing TabView.js with https://gist.github.com/dmhood/9e6e564d66277148c104497daa9a1b39 (if you want the logging I used).
2.  Navigate from the launch page to the "TabBar" page.
3.  Tab #2_1 will load initially.  Navigate via tabs to Tab # 1, and then back to Tab # 2.

The above steps will produce the following logging:
'mounting tabview', 'tabbar_0_tab2_1'
'mounting tabview', 'tab1_0_tab1_1'
'mounting tabview', 'tab2_0_tab2_1'

(additionally if you use my example you can see both instances of TabBar 2_1  generating logs).

I'm assuming this is due to a different scene key being created when 2_1 intially mounts vs when we navigate back to it via tab.  Is there a known workaround/solution for this?

ninjz:
Experiencing the same issue. For me this is also causing weird behaviours where touches are being captured and preventing scrolling on scroll views.

atlanteh:
+1

JunFenglee:
+1 happens to me too .
Adtionally, I change the navigation bar title  with "Actions.refresh("title":"changed Title")" after mounted the first tab. however, title change to default title after I switch to tab 2 and go back to tap 1.

rturk:
@dmhood I've glanced over your gist code, it looks quite different from the core code, specially considering that you're re-creating the tabs logic: 

`<Button onPress={() => { drawer.close(); Actions.tab1(); }}>Switch to tab1</Button>`

Why you need this? Can't you just use core tab logic and just redesign the UI part?

rturk:
FYI this works for me in production Apps

Please note the *   tabs={true}*  in the Tabs Scene, wrapped inside drawer menu

```
export default Actions.create(
  <Scene key="root">
      <Scene key="tabbar" component={DrawerMenu} type="reset" duration={1} initial={true} >
        <Scene 
               key="main"
               tabs={true}
               tabBarStyle={styles.tabBar}
               selectedIconStyle={styles.selectedIconStyle}
[...]

PR #1312, opened 2016-10-21
Restore 0.35 JUMP handling

### LINKED ISSUE #1270: The initial tab with scrollable views witch  can not be scrolled when tap between tabs and back to the initial tab
### Version
Tell us which versions you are using: 

- react-native-router-flux v3.36.0 
- react-native v0.32.0

I'd update to the latest version 3.36.0, but there comes a problem below. 

We have three tabs. The first one has a view pager witch used [react-native-viewpager](https://github.com/race604/react-native-viewpager) and a ListView witch from react-native.
The view pager is scrollable with horizontal direction while the ListView is scrollable with vertical direction.

The first tab is the initial tab in witch viwpager and ListView can both be scrolled, it works normal.  
But when you tap between tabs and back to the first one, both can be taped but not scrolled. The second tab has a view pager too without ListView, and it can be scrolled.
And I tried that let the second or the third tab be same as the first (initial) one then the other tabs is ok but only not the first (initial) tab. The tabs are all with same views though the initial one cannot be scolled. It seems that the views have not been updated as the content sure has been scrolled but not moved and you would tap the different content. Wish you can resolve this issue.Thanks.

 It's weird because that on react-native-router-flux v3.35.0 it works fine.

### Steps to reproduce
We have three tabs. The first one has a view pager witch used [react-native-viewpager](https://github.com/race604/react-native-viewpager) and a ListView witch from react-native
1.tap between tabs
2.back to the first one
3.neither Viewpager or ListView can be scrolled but can be taped.

wrozka:
Fixes #1270 #1311 and #1293 by restoring JUMP handling from v0.35

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -210,20 +210,8 @@ function inject(state, action, props, scenes) {
       };
     case ActionConst.JUMP: {
       assert(state.tabs, `Parent=${state.key} is not tab bar, jump action is not valid`);
-      /* TODO: recursive key search (for sub-tab scenes)*/
-      ind = state.children.findIndex(el => el.sceneKey === action.key);
-      // variant #1
-      // get target scene with data passed in Actions.SCENE_KEY(PARAMS)
-      const targetSceneWithData = getInitialState(props, scenes, ind, action);
-      // find same scene in state.children
-      const targetSceneInState = state.children[ind];
-      // update child scene from state.children with pased data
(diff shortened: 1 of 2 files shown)


czl1378:
Thanks for your PR!

haibinyu:
any new patch version for the fix?  IMO this is a critical issue.

aksonov:
@haibinyu Does master works well for you? Seems there are some other issues left...

PR #1370, opened 2016-11-12
Fix bug when specifying only tabIcon for TabBar

yasaichi:
Currently, TabBar isn't rendered when specifying only `tabIcon` without any `icon` for the child scenes.
As far as I looked over the code, I found out this bug was caused by the following one.

```js
state.children.filter(el => el.icon).length > 0
```
ref. https://github.com/aksonov/react-native-router-flux/blob/2b09f47740bb0f5829e8419d9ba6622aef20c42c/src/TabBar.js#L83

I think it is used for checking whether `Tabs` component has one or more `Icon` components.
However, actually it returns `false` In the above-mentioned case.

This PR fixes it.

### DIFF
--- a/src/TabBar.js
+++ b/src/TabBar.js
@@ -56,6 +56,13 @@ class TabBar extends Component {
       deepestExplicitValueForKey(state, 'hideTabBar') ||
       (this.props.hideOnChildTabs && deepestExplicitValueForKey(selected, 'tabs'));
 
+    const icons = state.children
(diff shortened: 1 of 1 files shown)


aksonov:
I'm really sorry about missing this PR. Unfortunately I have available time only recently and completed v4 beta based on React Navigation. Feel free to submit PRs for v4 now. I promise I will not miss them again.

Issue #1363, opened 2016-11-10
About hideNavBar property.


MechanicKim:
### Version
Tell us which versions you are using: 

- react-native-router-flux v3.36.0
- react-native v0.37.0

### Expected behaviour
As I know, default value of scene property 'hideNavBar' is false.

### Actual behaviour
but When I make hideNavBar true in specific scene, next other scenes are influenced.

### Steps to reproduce

1. Make three scenes and make hideNavBar of only second scene true.
2. If you reload your application, you can see that first page has a NavBar but, second and third don't have NavBar.

I've found a strange thing on DefaultRenderer.js
Two lines(237 ~ 238) are like this.

**const hideNavBar = deepestExplicitValueForKey(state, 'hideNavBar');
if (hideNavBar) {**

I fixed something(238 line) and I could solve this. like this.

**if (child.hideNavBar) {**

loug38:
Great fix! Thanks!

MechanicKim:
Thank you for your reply!

2016-11-16 5:07 GMT+09:00 Lou George <notifications@github.com>:

gavin-gmlab:
Similar issue #1330 . Is this going to be officially fixed?

MechanicKim:
Oh! Can I fix this issue?
I just wrote this issue. I didn't fix it.

sarovin:
@MechanicKim a PR is welcome ;)

Issue #1330, opened 2016-10-28
Nested Scenes with 2 successive hideNavBar Scenes causes all other Scenes  not to have Nav Bar


gavin-gmlab:
### Version

Tell us which versions you are using: 
- react-native-router-flux v3.35
- react-native v0.35

```
     <Router>
        <Scene key='drawer' component={NavigationDrawer} open={false}>
          <Scene key='drawerChildrenWrapper' navigationBarStyle={Styles.navBar} titleStyle={Styles.title} leftButtonIconStyle={Styles.leftButton} rightButtonTextStyle={Styles.rightButton}>
            <Scene initial key='splashScreen' component={SplashScreen} title='Components' hideNavBar='true' />
            <Scene key='mainLoginScreen' component={MainLoginScreen} title='Login' hideNavBar='true' /> 
            <Scene key='mainScreen' component={MainScreen} title='Home'/> <=== Nav bar no longer appears for all subsequent screens
... Other scenes
          </Scene>
        </Scene>
      </Router>
```
### Expected behaviour

Nav Bar should appear after the initial Splash and Login Screens
### Actual behaviour

Nav Bar never appears if I have 2 successive screens with hideNavBar='true'
### Steps to reproduce
1. Implement the code as above
2. Nav Bar on 3rd screen (mainScreen) does not appear.
3. Remove hideNavBar = true from the Login Scene
4. Everything works as per normal

peterkruty:
I have the same problem:

        <Scene 
          key="productintro"
          component={ProductIntro}
          hideNavBar="true"
          initial
        />

        <Scene
          key="homegym"
          component={HomeGym}
          hideNavBar={false}
          title="Home Gym"
        />

Second one does not have a NavBar.

peterkruty:
I made it work with:

Actions.homegym({hideNavBar: false})

gavin-gmlab:
@peterkruty, thanks - it kinda worked for me, but there were side effects.

I put the Actions.mainScreen({hideNavBar: false}) into my Button on the loginScreen - seems like this action is called automatically cause it automatically transitions straight to the mainScreen without me having to press anything (I am just doing UI work for now)

Also, there was a warning about setState(...): Cannot update during an existing State transition, which wasn't there before. There is no setState in my render() function for either screen. Could be from the Button component - not sure how it is implemented since I am using the one from 'react-native-elements'

shroukkhan:
[...]

PR #1406, opened 2016-11-22
Delete hideNavBar and replaced hideNavBar with child.hideNavBar

MechanicKim:
Please refer to #1363 

As I know, default value of scene property 'hideNavBar' is false.
So, I fixed two lines in a file 'Defaultrenderer.js'

Added information!
- To fix a problem about 'test' and 'jest', I updated  some react packages.

### DIFF
--- a/Example/__tests__/__snapshots__/Example-test.js.snap
+++ b/Example/__tests__/__snapshots__/Example-test.js.snap
@@ -1,34 +1,34 @@
 exports[`Example renders correctly 1`] = `
 <View
-  onLayout={[Function bound _onLayout]}
+  onLayout={[Function]}
   style={
     Array [
       Object {
         "backgroundColor": "transparent",
-        "flex": 1
+        "flex": 1,
       },
-      undefined
+      undefined,
     ]
   }>
   <View
     style={
       Object {
-        "flex": 1
+        "flex": 1,
       }
     }>
     <View
-      onMoveShouldSetResponder={[Function onMoveShouldSetResponder]}
-      onMoveShouldSetResponderCapture={[Function onMoveShouldSetResponderCapture]}
-      onResponderEnd={[Function onResponderEnd]}
-      onResponderGrant={[Function onResponderGrant]}
-      onResponderMove={[Function onResponderMove]}
-      onResponderReject={[Function onResponderReject]}
-      onResponderRelease={[Function onResponderRelease]}
-      onResponderStart={[Function onResponderStart]}
-      onResponderTerminate={[Function onResponderTerminate]}
-      onResponderTerminationRequest={[Function onResponderTerminationRequest]}
-      onStartShouldSetResponder={[Function onStartShouldSetResponder]}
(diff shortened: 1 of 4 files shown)


joenoon:
I'm a few versions behind and a bit out of the loop on this project, so I could be wrong here.  But removing deepestExplicitValueForKey seems to me like it would break a lot of functionality that used to be demoed in the Example app: Go to tabbar > push new scene > and then all of the variations of hideTabBar hideNavBar that persisted through additional scenes and correctly recoiled as they were popped off.  I would just urge someone more current on the state of the project to double check this. #1330 #1363

MechanicKim:
@joenoon 
I wanna help something as a person who uses this package.
what can I do for it?

joenoon:
@MechanicKim If I'm guessing, I think what you have is something like this:

```
- root hideNavBar=false
  - scene1 hideNavBar=false
  - scene2 hideNavBar=true
  - scene3 (unspecified hideNavBar)
```

And when you go Actions.scene3(), you notice it thinks `hideNavBar == true`

If that accurately describes the issue, that was how I intended hideNavBar and hideTabBar to work - they are inherited values.  The `deepestExplicitValueForKey` function walks the scene tree to get the inherited value when it isn't explicitly defined.  Those props are special/different than other props to get the navigation behavior in the "push new scene" demo.  I would compare it to how the ios UINavigationController behaves as you push and pop scenes while setting navigationBarHidden and hidesBottomBarWhenPushed as you go.

So I guess a few things:

1) Am I describing the issue right?

2) If so, it would not be a bug, so would better documentation resolve this? (either explicitly set hideNavBar on scenes, or at least set hideNavBar=false on any scene following a hideNavBar=true so it goes back to false).

3) Or is this not a behavior people want?  For my app this functionality is critical and everywhere, but I can work around it if the core library wants to remove it.

Also, I mentioned before I'm a few versions behind, so I'm kind of jogging my memory on this, and I can't vouch that this functionality hasn't been broken in the latest versions that I haven't seen yet.

MechanicKim:
First, I'm sorry. I might not know the intention of this react native package well.
To build our application, we are using this package. I just felt strange about using navigation bar.
For example, some html form tags have their default prop values. If we don't change it, they keep the default value. I have been used to this way.

joenoon:
@MechanicKim no need to be sorry!  What do you think about the rest?  Does my explanation solve your issue or is there still something that doesn't work for your app?

MechanicKim:
Yes! I totally understood your explanation that is so clear. ^^
No problem in our application. I just felt a little strange.
If I can contribute to this project in anyway, I hope to do it.
Thank you so much!

Have a nice day.

2016-11-25 1:28 GMT+09:00 Joe Noon <notifications@github.com>:

charpeni:
Thank you @MechanicKim, your time is greatly appreciated.

I'll close this PR since it seems resolved, if you think it shouldn't be closed, please let me know.

PR #1415, opened 2016-11-23
Temporarily disable jest testing on Example project

charpeni:
There is an issue with jest testing, so I disabled it until I have time to fix this.

Parallelism have been set to 2 containers.


sarovin:
@charpeni Now we have another error in circleci, Can you fix it, please?

PR #1418, opened 2016-11-24
add additional enum value for "direction"

dannycochran:
I think the "animation" and "duration" props are identical, such that one should probably be deprecated.


bang88:
Thanks @daemonchen ,but the tests failed.

charpeni:
@dannycochran Are you up to rebase on master?

PR #1425, opened 2016-11-26
Update Eslint and Airbnb rules

MechanicKim:
1. I updated eslint packages(5 packages)
2. I fixed style. The revised details are as follows.
- Missing trailing comma
- Expected parentheses around arrow function argument having a body with curly brace
- Object properties must go on a new line if they aren't all on the same line
- Unary word operator 'typeof' must be followed by whitespace
- Expected indentation
- Unexpected parentheses around single function argument having a body with no curly braces

Total 133 problems, 53 problems were solved, 80 problems remain.

Because I thought the rest of the problems are touch sensitive, I didn’t fix them.

I didn't pass 'Circle CI'. So, I added some rules.

**disabled**
1. class-method-use-this
2. import/no-name-as-default
- it says name ‘getInitialState’ is wrong but I couldn’t change this name.
3. react/forbid-prop-types

**modified**
1. react/jsx-filename-extension
- To allow ‘JSX in js extension’
2. import/no-extraneous-dependencies
- To allow ‘react package in devDependencies’
3. no-bitwise
- To allow ‘^’
4. no-restricted-syntax
-  To exclude ‘forOfStatement’

Also, I fixed a few more style problems.
As a result, I fixed all style problems.


charpeni:
Amazing, thank you @MechanicKim!

I'll review this soon.

.eslintrc.js
```diff
@@ -5,6 +5,9 @@ module.exports = {
   ],
   parser: 'babel-eslint',
   rules: {
+    'class-methods-use-this': 0,
```
charpeni: `class-methods-use-this` should be enabled. 

If a class method does not use `this`, it can safely be made a static function.

.eslintrc.js
```diff
@@ -5,6 +5,9 @@ module.exports = {
...
+    'import/no-named-as-default': 0,
```
charpeni: `import/no-named-as-default` should be enabled, it's likely misleading.

charpeni:
In addition, these changes will need to be tested. 

If someone can confirm that everything works well, unfortunately I can not build on mobile for now.

MechanicKim:
Okay I'll do that you commented.
Thank you for your review and comments. :-)

2016-12-01 10:19 GMT+09:00 Nicolas Charpentier <notifications@github.com>:

MechanicKim:
After fixing commented and testing, I confirmed that testing was successful!
But, I failed 'npm run jest'. I tried various version of react packages but I got an same error.
So, I registed an issue on Facebook/jest(https://github.com/facebook/jest/issues/2208)

MechanicKim:
@charpeni I resolved your request. :-)

src/DefaultRenderer.js
```diff
@@ -135,21 +135,21 @@ export default class DefaultRenderer extends Component {
     }
   }
 
-  getPanHandlers(direction, props) {
-    return direction === 'vertical' ?
+  getPanHandlers = (direction, props) => (
```
charpeni: Unfortunately, the proposal for public class fields & static properties is only at Stage 2.

It should be a static function with the keyword `static`, not an arrow function.

Same thing on other methods.

MechanicKim:
Okay! I'll change and push again! Thank you :-)

2016-12-05 8:24 GMT+09:00 Nicolas Charpentier <notifications@github.com>:

MechanicKim:
@charpeni I resolved your request. :-)

charpeni:
One last thing before merging, everything works fine on both platforms?

MechanicKim:
I tested example project on both platforms.
When I go to the Login2, I get an error because this.props.data is not string.
this data is object.
I will test it one more and I'll let you know when I finish it.

Thanks!

MechanicKim:
@charpeni I found that that is not a problem.
When I go to Login page, send data to Login page.({data:"Custom data", title:"Custom title" })
But, when I go to Login2 page, don't send data to Login2 page. So, red screen is right.
If Xcode Build Configuration is 'Release', red screen is not shown.

I tested iOS and Android project. :-)

charpeni:
There was also a red screen before ?

I'd like to merge this pull request, unfortunately I can not introduce breaking changes, that would be problematic for developers.

MechanicKim:
As you know, when I press 'Go to Login page' in Launch.js, data is transmitted.
`<Button onPress={()=>Actions.login({data:"Custom data", title:"Custom title" })}>Go to Login page</Button>`

when I press 'Login 2' in Login.js, nothing is transmitted.
`<Button onPress={Actions.loginModal2}>Login 2</Button>`
But, In Login2.js, it tries to access to data(this.props.data).
`<Text>Login2 page: {this.props.data}</Text>`
'this.props.data' is not undefined.
So, I think this error is intended.

charpeni:
@MechanicKim I tested `master` branch on Android and everything works fine without red screen on Login 2. If you could fix this it would be appreciated so we can finally merge this. Thank you!

MechanicKim:
@charpeni Today, I downloaded master branch and tested Example project.
I got same red screen. Please check 'build configuration' in your Xcode scheme.
If this value is 'Release', you can't show red screen!

charpeni:
@MechanicKim I can only test on Android.

MechanicKim:
@charpeni 
I tested master branch and I confirmed that the results are same each other. :-)

MechanicKim:
@charpeni I fixed Example project.
I think you might able to understand my change.

charpeni:
Thank you for your time :)

PR #1488, opened 2016-12-19
Update API_CONFIGURATION.md

javiercr:
Added missing [`titleWrapperStyle`](https://github.com/aksonov/react-native-router-flux/blob/master/src/NavBar.js#L160) prop.


cridenour:
Thanks!

PR #1512, opened 2017-01-02
New tabBar features

davidshore:
I added the possibility to have a floating tab bar, that floats above the tabbed view. This is very important when your tab bar has an irregular shape. Like in my case in shape of a cloud and I want the tabbed view to appear behind the cloud.

I also added the possibility to style the background image of the tab bar.


aksonov:
@davidshore I'm really sorry about missing this PR - I unsubscribed to PR once and forgot to re-subscribe again later. Unfortunately I have available time only recently and completed v4 beta based on React Navigation. Feel free to submit PRs for v4 now. I promise I will not miss them again.

PR #1524, opened 2017-01-08
(typings) update APIs, remove duplicate functions, make it react-redux friendly

pocesar:
* Made the typings more strict in some cases
* Added new props that were added recently
* Fixed some inconsistencies with the docs
* Scene now extends from TouchableWithoutFeedbackProperties (so no more non-declared onPress for example for the new API)
* Removed duplicated, non documented interfaces from `RouterProps`
* Made it `react-redux` friendly (specially when using `connect` for `Switch`, for example, but works for any other RNRF components)
* Made classes extendable friendly (it can be augmented with ownProps, as React.Component dictates)
* Typings now follow tslint defined by DefinitelyTyped as best-practices
* Declarations can now be extended globally in a proper way (it wasn't possible because the current typings are incorrect, they doesn't export anything and act like a global declaration)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ node_modules/
 npm-debug.log
 
 .vscode/
+tsconfig.json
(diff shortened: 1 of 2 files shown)


aksonov:
@pocesar  I'm really sorry about missing this PR - I unsubscribed to PR once and forgot to re-subscribe again later. Unfortunately I have available time only recently and completed v4 beta based on React Navigation. Feel free to submit PRs for v4 now. I promise I will not miss them again.

PR #1525, opened 2017-01-08
Fix inconsistent navbar button placement

jakecraige:
The back button / left button were being set to different vertical positions.
This updates those to be consistent for left and right.


Cody-G-G:
Maybe also add the missing `paddingLeft: 6,` in barLeftButtonText, since that's also an inconsistency?

PR #1472, opened 2016-12-13
Pass props down to child components

DonnieWest:
Fixes #1434

Issue #1434, opened 2016-11-28
Props not passed from tab scene to sibling scene


ssomnoremac:
### Version
Tell us which versions you are using: 

- react-native-router-flux v3.36
- react-native v0.37

### Expected behaviour
object props are passed to scene like so 
```
Actions.myScene({foo: {bar:"baz"}})
```
### Actual behaviour
the props are not showing up in myScene

there is another issue that was closed #1390 for no reason. There were many breaking changes in RN36,37. Anyone else seeing their entire app break suddenly?

sarovin:
Can you show an example?

ssomnoremac:
<Scene key="worker" getSceneStyle={getSceneStyle} component={DrawerComponent} open={false}>
         <Scene key="workerTabs" getSceneStyle={getSceneStyle} tabBarStyle={tabBarStyle} tabs={true} >
               <Scene key='workerHome' component={ WorkerHome } title="Home" />
               <Scene key='workerShiftDetails' component={ WorkerShiftDetails } title="Home" hideTabBar={true} />

In `WorkerHome` navigating like so `Actions.WorkerShiftDetails({foo:"Bar"})

Mind you I am using both Drawer and Tabs and all of this is a subScene so perhaps one of those 3 things could be to blame?

sarovin:
If you don't use drawer or tabs, it works?

ssomnoremac:
Yeah, it's working in other parts of the app. And I tried to pass just a string. No luck, just nothing is passed along in the props. Using Redux `mapStateToProps` in the next scene or not, nothing gets through.

sarovin:
Sorry @ssomnoremac, i do not know...I can try to investigate when I have time.

DonnieWest:
Ran into this issue at work. Unsure if #1472 is the best solution, but it works for us

MechanicKim:
I tested passing value(props) on Example project.
I have no problem.
I pass the value like this.
`{data:"Custom data", title:"Custom title", child: {data:"Child data"}}`
In next page, I can get value from this.props.child.data.

DonnieWest:
@MechanicKim since I can't share my at work project, I'll try to get a minimal project setup that demonstrates the bug based on the example project.

The props only don't pass when the parent scene has tabs set and you are passing the props to a sibling like in the example from @ssomnoremac

warrenronsiek:
[...]

PR #1549, opened 2017-01-16
pass props

indatawetrust:
The option of sending props to children's scenes was added.

When _passProps_ is true props are sent.

`<Scene key="user" component={ConnectUser} title="User" hideNavBar="true" passProps={true}/>`


aksonov:
Some tests fail, could you check?

DonnieWest:
@aksonov this is another fix for #1434 and shouldn't be needed after #1472 was merged. The big difference is that this makes passing props to the child components optional and I think most people who are using this library expect those props to be passed without additional configuration

PR #1552, opened 2017-01-16
Override back button if left button provided

superandrew213:


superandrew213:
@aksonov can we take this in? It's only a simple change which was intended to be implemented like this.

aksonov:
@superandrew213 It will brake existing apps. Please introduce special setting like overrideBackButton, document it and re-send PR, thanks.

superandrew213:
@aksonov are you sure that it will break existing apps?

At the moment if you set the left button, the back button overrides it. So basically you can't set the left button at all at the moment. The back button is set by default and overrides the left button if it is set.

```
{renderBackButton(navProps) || renderLeftButton(navProps)}
```

The change would only break apps that have set the left button (mistakenly). Because it is not working at the moment and would only show the back button. After the update the left button would override the back button.

aksonov:
Let me explain. For my app, i want to see “<“ button anywhere. Only for first scenes in the stack i want to see ‘left’ button (i.e. hamburger button).

superandrew213:
Ah ok, I got you.

aksonov:
Will you re-work this ticket with adding new property or I could close it?

Issue #384, opened 2016-03-24
Double tapping on link creates a double view


AlexDM0:
Hi,

When using the example, if you click on a link twice quickly (so before the new view has finished the animation) it will trigger a duplicate of the required view to slide in. Is this something we need to check in our reducer implementation or is it a bug?

Very ugly workaround:
```js
let startTime = new Date().valueOf();
let lastTime = new Date().valueOf();

const reducerCreate = params=>{
    const defaultReducer = Reducer(params);
    return (state, action)=>{
        // startTime so the application has booted, 250 for animation duration.
        if (new Date().valueOf() - lastTime < 250 && new Date().valueOf() - startTime > 1000) {
            return state;
        }
        lastTime = new Date().valueOf();
        return defaultReducer(state, action);
    }
};
```

I have a few big buttons that the user can click on and if they accidentally tap twice, I dont want to load 2 cards of the next view.

Regards

evollu:
UIExplorer example has the same issue. I'm doing the same workaround

slamus:
You should use the reducer to store the currentRoute somewhere in your app.
Then you check if the currentRoute is already set and prevent the route to be triggered twice.

AlexDM0:
I had that before but then the example didn't work as it is calling a reference to itself. In production that wouldn't be the case so there your solution would indeed be better.

aksonov:
Check 3.2.2, reducer doesn't change state for push the same route again.

wkwyatt:
This issue persist when there is an object passed to the route.  
i.e. Actions.newScene({ params: { id: 5, name: "newRoute" } })

What should be done if this is the issue?

PR #1561, opened 2017-01-19
Fix checkPropertiesEqual

barakcoh:
Hi

Thought I'd take a stab at fixing this as per the `WARN` comment above it. 

In our case, we pass quite a few props in navigation calls, including arrays and functions. The `!==` check fails for those, allowing for the double-tap effect mentioned  in https://github.com/aksonov/react-native-router-flux/issues/384

The proposed solution is to use lodash's [isEqual](https://lodash.com/docs/4.17.4#isEqual) which provides a solid equality check that should suffice for most cases. `isEqual` is bundled separately to prevent bloating RNRF.

Additionally to `isEqual` I also opted to exclude checking functions altogether, as those are usually generated inline as part of the navigation call, failing `isEqual` as well.

Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     }
   ],
   "dependencies": {
+    "lodash.isequal": "^4.5.0",
     "react-addons-pure-render-mixin": "15.0.2",
     "react-native-experimental-navigation": "0.26.x",
     "react-native-tabs": "^1.0.9",
--- a/src/NavBar.js
+++ b/src/NavBar.js
@@ -52,7 +52,7 @@ const styles = StyleSheet.create({
     position: 'absolute',
     ...Platform.select({
       ios: {
-        top: 20,
+        top: 15,
       },
       android: {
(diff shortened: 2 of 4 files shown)


Naoto-Ida:
This is great. I have the same issue of double calls to Actions which cause two pushes to happen.

barakcoh:
@aksonov are you interested in merging my fix? I'll re-open the PR w/ the just 1st [commit](https://github.com/aksonov/react-native-router-flux/pull/1561/commits/33d5ec3a203d5fb8cbb975d7ae66c04d4a33c54b) if you are

aksonov:
@barakcoh Yes, I'm interested

aksonov:
Could you please resolve conflicts and re-send PR, I've fixed the build

barakcoh:
Closing in favor of https://github.com/aksonov/react-native-router-flux/pull/1682

Issue #1216, opened 2016-09-24
Missing properties in index.d.ts for SceneProps


joewood:
### Version

Tell us which versions you are using: 
- react-native-router-flux v3.35.0
- react-native v0.32

Some properties are missing in index.d.ts for SceneProps :
- `open?:boolean` - used when the `Scene` is holding a Drawer
- `icon?:JSX.Element` - used when the `Scene` is a tab in a tabbar.

PR #1576, opened 2017-01-24
#1216 - Add `open` prop to TS declaration

BrendanBerkley:
This is messing with my builds, and it's tough to work around because I don't know how (don't know if it's possible?) to assign a JSX prop to the any type. Hope I put this in the right spot in the file!


BrendanBerkley:
☝️ Pretty confident that build error isn't the fault of this PR. Any way to re-run it?

charpeni:
@BrendanBerkley Done, thank you!

Issue #1578, opened 2017-01-24
Pop action difference between Android and iOS


skoob13:
### Version

- react-native-router-flux v3.37.0
- react-native v0.39.2

### Expected behaviour

On multiple Actions.pop() calls when there are no routes to pop shouldn't throw an error "You are already in the root scene".

### Actual behaviour

On iOS the multiple calls of the Actions.pop() won't be an error. On Android the same code throws an error "***You are already in the root scene***".

### Steps to reproduce

1. Run Android simulator or device
2. Call multiple times Actions.pop() when there is more than one scene in stack
3. Error

Is there any solution to handle this error on Android?

skoob13:
Seems like in 0ef1f7a19ca30d99e1892595da5be67a15c8fc22 was firstly introduced this error on Android and removing lines from [121 to 123](https://github.com/aksonov/react-native-router-flux/blob/master/src/Reducer.js#L121) fixes this issue.

afilp:
I confirm that this behavior has been a problem. Can someone tell us if there is a reason for this warning? I mean why it was added on the first place.

What happens when this code is omitted, I assume the app exits?

Swordsman-Inaction:
Can you check PRs #820 #1049 and #1451 please, that's why I did this. It seems like you got the same issue with #1165, after #1451, this issue has been fixed.

PR #1451, opened 2016-12-04
Fix android back error

Swordsman-Inaction:
The old way to implement android back handler will raise an issue described in #1165

This PR will fix this by create a special action just for android back action.

@cuitianze @Micjoyce Can you test this a little bit, to make sure this does fix the issue on your project?

PR #1580, opened 2017-01-25
Fix Android pop action error

skoob13:
Related to #1578


aksonov:
See https://github.com/aksonov/react-native-router-flux/pull/1451

PR #1584, opened 2017-01-27
Changes drawer button side according with drawer.props.side

lucianomlima:
Uses the side drawer prop to decide where to render the drawer button.
drawer.props.side == 'left', renderLeftButton (default).
drawer.props.side == 'right', renderRightButton.


Cody-G-G:
This is great, been looking into how I can get this done! Hoping for a merge soon 👍

PR #1614, opened 2017-02-08
Documentation for drawer (side menu) integration

jeongmincha:
- Add class name 'NavigationDrawer' instead of omitted class name.
(The name for the component class was **gone**! It is a critical problem.)
- Split code sections for more readability


jeongmincha:
@aksonov @adrianmcli @brien-crean @dragfire Would you please check this MR? It's one about documentation. I think you don't need have much time to decide whether it's necessary or not. 
Thank you in advance. I also appreciated this good library for react-native.

Issue #1174, opened 2016-09-15
Transitioning out of scene with does not use custom animationStyle


peritus:
### Version

Tell us which versions you are using: 
- react-native-router-flux v3.35.0
- react-native v0.33
### Expected behaviour

When using a scene with custom animationStyle prop, e.g.

```
<Scene key='Overview' component={Overview} panHandlers={null} animationStyle={myCustomHorizontal} />
```

I expect that animationStyle interpolation function to be used for transitioning in _and_ out of the navigation card.
### Actual behaviour

Only the transition to the card is using the custom animationStyle interpolation function, the transition out is using the default behaviour. This used to work in v3.34 (we downgraded to that version for now).
### Steps to solve (maybe?)

Revert https://github.com/aksonov/react-native-router-flux/commit/04353d951ff88c35284c7e5baf73d19f79843a8e (Part of #1058)

wachunei:
I'm new to RN and I just found out how to set this for scenes and it is amazing. But my transitions back to previous scene are using default `NavigationCardStack` functions, which mess with the UX of my app.

@aksonov please consider this bug to be important for an update 🙏🏼

wachunei:
@peritus I realized it is not using default behavior, but using animationStyle of previous scene in the stack.

Let me explain: consider a base scene (A), on top of that I present a new scene (B) with custom animation style, as you explained, this scene will use that animation to push itself into the stack and when using panHandlers to animate a transition back, but when using `pop()` it will use default animation.
Let's say I push another scene on top of that (C) with another custom animation: it will use such animation to push itself and transition back with gestures, but at the time of using `Actions.pop` it uses B's animation.
That's why I conclude B is using A's animation to pop, which happens to be the default one.

I wish any collaborator or @aksonov would take a look into this 😕

Jlexyc:
Got the same issue.. Did a fix in router-flux to avoid it. DefaultRenderer.js in "RenderCard" method reads animationStyle from the wrong scene...

wachunei:
@Jlexyc Would you elaborate, please? Should `animationStyle` be read from `props.scene.navigationState` instead?

Jlexyc:
@wachunei i'm using:

` let panHandlers = selected.panHandlers;
    let animationStyle = props.scene.navigationState.animationStyle`

But there still an issue with transition. Like animation is called for both screens with different index. So i'm covering disappearing from old screen and appearing from new one.

[...]

PR #1618, opened 2017-02-09
Fix for panGestures and Animation

Jlexyc:
ISSUE:

AnimationStyle was taken from the new Scene even for POP animation. So different transition on PUSH in POP.

getPanHandlers - was taken from "PARENT" only and there was no way to override it in child (or disable).


wachunei:
This closes #1174

PR #1682, opened 2017-03-07
Fix checkPropertiesEqual

barakcoh:
Supersedes #1561

In our case, we pass quite a few props in navigation calls, including arrays and functions. The !== check fails for those, allowing for the double-tap effect mentioned in #384

The proposed solution is to use lodash's isEqual which provides a solid equality check that should suffice for most cases. isEqual is bundled separately to prevent bloating RNRF.

Additionally to isEqual I also opted to exclude checking functions altogether, as those are usually generated inline as part of the navigation call, failing isEqual as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     }
   ],
   "dependencies": {
+    "lodash.isequal": "^4.5.0",
     "react-addons-pure-render-mixin": "15.0.2",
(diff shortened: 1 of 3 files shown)


barakcoh:
@aksonov here's the updated PR

Issue #1508, opened 2016-12-30
Native module cannot be null


Chren:
### Version
Tell us which versions you are using: 

- react-native-router-flux v3.37.0
- react-native v0.39.2

### Expected behaviour
no error

### Actual behaviour
Native module cannot be null

NativigationRootContainer.js 15

### Steps to reproduce

1. import {Scene, Router} from 'react-native-router-flux';
2. no step2
3.

![simulator screen shot 2016 12 30 10 57 27](https://cloud.githubusercontent.com/assets/349182/21558965/e037ba4e-ce7d-11e6-8088-9c7ef8f896f1.png)

aiyidu:
solution:
You need to add RCTLinkingIOS in your pod file, which the documentation didn't mention.

wcandillon:
@aiyidu Thank you for the solution.

Chren:
It works! I am going to close this issue.

affibox:
I actually added the RCTLinkingIOS to my pod file but I'm still getting the same error. Nothing changed.

pgonzalez-santiago:
Same error here trying to run the Example. :/

# Get the code
git clone https://github.com/aksonov/react-native-router-flux.git
cd react-native-router-flux/Example

# Install dependencies
yarn

# Run it
react-native run-ios

aksonov:
RNRF doesn't use any native modules, check stack trace.

pgonzalez-santiago:
ok. Thank you. The problem was I had another RN project running on background.

PR #1714, opened 2017-03-16
Update README.md

erlendev:
Based on this issue https://github.com/aksonov/react-native-router-flux/issues/1508 and my experience from setting things up, this has to be added to the installation.


aksonov:
Sorry, this component doesn't use any native modules. I've just updated demo with 'react-native init Example' and it works without any additional steps...

PR #1742, opened 2017-03-27
update documentation for nested routing

cosmos-sajal:
#### What
Document updation for nested routing between scenes, i.e. from one scene nested in one subroute to other scene nested in other subroute.

### Why
To update the documentation as it will be helpful to others as well.


cosmos-sajal:
@aksonov can you check and verify?

Issue #1758, opened 2017-04-03
Back button disappears when I switch tabs


shwetachoudhary06:
### Version
Tell us which versions you are using: 

- react-native-router-flux v3.38.0
- react-native v0.42.0

Following are my scenes

```
<Scene key="tabbar" tabs={true}>
     <Scene key="tab1" title="Tab 1" icon={TabIcon}>
       <Scene key="recommendations" component={RecommendationsScene} title="Reccomendations" />
     </Scene>
     <Scene key="tab2" title="Tab 2" icon={TabIcon}>
       <Scene key="graph" component={GraphScene} title="Graph" />
     </Scene>
     <Scene key="tab3" title="Tab 3" icon={TabIcon}>
       <Scene key="editMetric" component={EditMetricScene} title="Edit Metric" />
     </Scene>
   </Scene>
```

The tab screen shows everything but when I switch between tabs, the Back button disappears.
It seems the Nav Bar is getting completely refreshed when switching tabs.

JCDJulian:
Having the same issue. It would be nice to get this checked out.

OhaiBBQ:
+1

@JCDJulian @shwetachoudhary06 Do either of you know if this behavior recently started occurring? Was it working in past releases?

OhaiBBQ:
I've been digging into this. My scene schema works with v3.36.0. It seems that #1312 reverted functionality that made this work, since there were some other bugs that were introduced by #1057.

I'll see if I can figure out what issues there were with #1057 & add this functionality back to RNRF.

JCDJulian:
This is my first time using RNRF so I don't have anything to compare it to. Thanks for looking into it!

raphh11:
Having the same issue using `react-native-router-flux@3.35.0`

OhaiBBQ:
Just submitted a PR, hopefully it gets merged soon. Waiting for feedback there.

PR #1775, opened 2017-04-07
Maintain tab parentIndex when jumping tabs

OhaiBBQ:
Hi @aksonov & all, this pull request fixes issue #1758.

I'm new to RNRF, so I don't know the full history, but: this functionality appears to not exist in 3.35 and was introduced in 3.36. I believe this regression was introduced by 6526529 as this functionality was broken 3.37+. 

When performing a JUMP action, we were replace the state of the `tabs` route child, thus replacing the special `parentIndex` attribute that is introduced to the children when a `tabs` route is PUSHed.

The observed user behavior is that the navigation bar's back button will disappear after switching tabs on a PUSHed `tabs` route.

Please let me know if you think there is test coverage missing, or would prefer a different implementation. The way I fixed it feels odd, as it seems to be mixing concerns of the `Reducer` and the `State` functionality.

Thanks,
Dan

### DIFF
--- a/src/Reducer.js
+++ b/src/Reducer.js
@@ -230,7 +230,7 @@ function inject(state, action, props, scenes) {
         props,
         scenes,
         state.index,
-        action,
+        { ...action, parentIndex: state.children[ind].parentIndex },
       );
 
       return { ...state, index: ind };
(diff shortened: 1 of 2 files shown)


OhaiBBQ:
Tests pass locally, I'm not sure what the Jest error on Travis is.

OhaiBBQ:
thanks for the merge, @aksonov!

Issue #1809, opened 2017-04-21
 Duplicate module name: react-native-packager


itinance:
### Version
Tell us which versions you are using: 

- react-native-router-flux v3.38.1
- react-native v0.43.3

### Expected behaviour

everything works fine

### Actual behaviour

error:

> Failed to build DependencyGraph: @providesModule naming collision:
>   Duplicate module name: react-native-packager
>   Paths: $$$$/node_modules/react-native-router-flux/node_modules/react-native/packager/package.json collides with $$$$$$/node_modules/react-native/packager/package.json
> 
> This error is caused by a @providesModule declaration with the same name across two different files.
> Error: @providesModule naming collision:
>   Duplicate module name: react-native-packager
>   Paths: $$$$$$$$/node_modules/react-native-router-flux/node_modules/react-native/packager/package.json collides with $$$$$$$/node_modules/react-native/packager/package.json
> 

### Steps to reproduce

1. react-native init
2. yarn add react-native-router-flux
3. require this module
4. build/run the application

ChristianTucker:
In the mean-time you can remove react-native from the node_modules folder in react-native-router-flux. however ultimately the example dependencies should be removed from the module and added to the examples package.json

itinance:
Thx @ChristianTucker This helped!

itinance:
PR sent

leyuan:
I have tried to remove them and move them to `devDependencies` and `peerDependencies`, but I got the following errors
```
 FAIL  __tests__/xxx-test.js
  ● Test suite failed to run

    ENOENT: no such file or directory, stat '{projectDir}/node_modules/react-native-router-flux/node_modules/react-native/Libraries/Core/ErrorUtils.js'
        at Error (native)

      at Object.fs.statSync (fs.js:1000:11)
      at Object.statSync (node_modules/graceful-fs/polyfills.js:297:22)
      at Object.<anonymous> (node_modules/react-native/jest/setup.js:64:532)

```

Am I missing something here?

leyuan:
Also is `React` a needed `dependency`? It caused me [react-ref-owner-error](https://facebook.github.io/react/warnings/refs-must-have-owner.html)

pckz:
1. ```npm i react-native-router-flux@3.38.0 --save```
2. ```npm cache clean```
3. ```react-native run-ios```

wuhenwuya:
+1

PR #1808, opened 2017-04-21
Move react deps to peerDeps, fixes #1803

stensrud:

PR #1812, opened 2017-04-22
removed react-native-dependency as it will result in duplicate-module…

itinance:
This will fix https://github.com/aksonov/react-native-router-flux/issues/1809

For testing you can use the following line dependencies of package.json:

```
	"dependencies": {
		"react": "16.0.0-alpha.6",
		"react-native": "0.43.4",
		"react-native-router-flux": "git:github.com/itinance/react-native-router-flux#fc1b8261144fe5f3132ac4940dd794ac06e9f351"
	},

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "react": "^15.4.2",
     "react-addons-pure-render-mixin": "^15.4.2",
     "react-dom": "^15.4.2",
-    "react-native": "^0.41.2",
     "react-native-experimental-navigation": "0.26.x",
     "react-native-tabs": "^1.0.9",
     "react-static-container": "1.0.1"
--- a/package.json
+++ b/package.json
@@ -67,7 +67,8 @@
     "jest": "19.0.1",
     "react-test-renderer": "15.4.2",
     "react": "^15.4.2",
-    "react-native": "^0.41.2"
+    "react-native": "^0.41.2",
+    "react-native-button": "^1.8.2"
   },
   "jest": {
     "preset": "react-native"
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
   ],
   "dependencies": {
     "lodash.isequal": "^4.5.0",
-    "react": "^15.4.2",
     "react-addons-pure-render-mixin": "^15.4.2",
     "react-dom": "^15.4.2",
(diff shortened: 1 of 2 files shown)


byennen:
I recommend moving `react-native` to "devDependencies" instead of removing it totally.

itinance:
ok, thx for the hint. Did it.

radko93:
It's the same case with `react` dependency.

leyuan:
Hey @itinance it seems like the `yarn.lock` file needs update too, thanks for the good work!

itinance:
ah good to know. thanks :) i wasn't sure if i should add the yarn.lock or not

leyuan:
@byennen why we want to keep it as a `devDependency`? I don't think it's being used anywhere..

In #1809 
> In the mean-time you can remove react-native from the node_modules folder in react-native-router-flux. however ultimately the example dependencies should be removed from the module and added to the examples package.json

If we want to keep it we should keep it as [Peer Dependency](https://yarnpkg.com/lang/en/docs/dependency-types/)

@itinance yeah sorry I didn't notice it's in dev dep right now, I mean if we want to totally remove it, we also need to remove it from lock file.. hmm not sure for now!

itinance:
@leyuan why not :) i thought it would be used for tests via jest and the _react-test-renderer_. But to be honest i am not familiar with that

i will try another commit with removing react-dependency at all.

leyuan:
@itinance if we keep it as a dev dep, it will still get installed into the node_modules under `react-native-router-flux` right? I think we will have two versions of react-native again..

itinance:
But anyways, the checks fail with 

> Module ./jest/setup.js in the setupFiles option was not found.

Has this something to do with the issue that we are discussing here? i think not.

Mikaila94:
Any update on this?

ohtangza:
I guess this should be moved to peerDependency as submitted here: https://github.com/aksonov/react-native-router-flux/pull/1808

msqar:
anyone knows how to fix this? i still can't use this dependency :(

ohtangza:
Sadly, I switched back to the official react navigation, which works very good so far for me.

aksonov:
Should be fixed now. Also v4 is released (based on new React Navigation)

Issue #1433, opened 2016-11-28
Back button appearing in root of navbar scenario


AlexDM0:
### Version
Tell us which versions you are using: 

- react-native-router-flux v3.36.0
- react-native v0.37.0

### Expected behaviour

If I move from a scene outside the navbar to one inside the tab bar (thing at the bottom) with type RESET, there is a back button in the navbar (thing at the top)

Scenario (pseudo)
```js
<Router ...>
  <Scene key="Root" >
    <Scene key="s1" component={...} hideNavBar={false} direction="vertical" />
    <Scene key="s2" component={...} hideNavBar={false}  />
    <Scene key="s3" tabs={true} hideNavBar={true}  type="reset" initial={false}>
      <Scene key="overview" tabTitle="..." icon={...} iconString="..." >
        <Scene key="ov" component={...} hideNavBar={...} />
      </Scene>
      <Scene key="settings" tabTitle="..." icon={...} iconString="ios-cog"  initial={...} >
        <Scene key="s" component={...} hideNavBar={...} title="."/>
      </Scene>
    </Scene>
  </Scene>
</Router>
```

I start in scene s1, then do Actions.s3({type:'reset'}), then click the tab icon at the bottom for the settings tab and tadaa there is a back button on the top. The log says the type from overview to settings is JUMP and 2x focus. If I change the initial in s3 to true, this behaviour does not occur.

### Expected behaviour

The back button should not be there

Cheers

sarovin:
Can you try this fix? #1386

AlexDM0:
I'll try it once I have time. Our internal deadlines are quite pressing at the moment so I worked around it by using my own navbar component in this view. 

Regards

sarovin:
no problem @AlexDM0

AlexDM0:
Are there any plans to refractor the usage of the navbar? I'd rather have a navbar object passed to the scene so the scene is fully responsible for the content and style of the navbar instead of the awkward separation between the router/scene definition and the scene itself. Dynamic titles, a bit more complicated buttons etc are all annoying to define outside of the scene. They can depend on an internal state or processing of the props which is done in the scene. 

Regards

PR #1838, opened 2017-05-04
fix "Back button appearing in root of navbar scenario" #1433

twksos:
We meet the issue in our project and this fix works for us


twksos:
@zxcpoiu , @MechanicKim , @aksonov 
please let me know if there is any concern on merge this one.

PR #1900, opened 2017-05-26
Add a down chevron icon for scenes with a vertical direction

markdaws:
With this change, when the user specifies direction="vertical" on a scene, instead of displaying a back chevron we display a down chevron, since when the user clicks on the back button the scene transitions down to the bottom of the screen not horizontally.


aksonov:
Sorry, but I believe it is out of scope for this component, you always could customize it from your app.

Issue #2023, opened 2017-07-13
Navigation 'back' button won't render on Android for scenes with 'back' attribute.


quarryman:
Issue can be reproduced on Example project.

Given we have scene with 'back' attribute passed
`<Scene key="echo" back clone component={EchoView} getTitle={({navigation}) => navigation.state.key}/>`

Router renders 'back' button for iOS (top left corner)

<img width="316" alt="screen shot 2017-07-13 at 11 07 43 am" src="https://user-images.githubusercontent.com/1093040/28156958-7bdf5e2e-67bd-11e7-8c42-9e330382aaa5.png">

And doesn't render back button for Android (back button missing)

![andoid-back-button-missing](https://user-images.githubusercontent.com/1093040/28157365-ed51d8ba-67be-11e7-9d9b-4cc78d11da77.png)

Initially found on other project and reproduced with Example to consider it bug.

aksonov:
@quarryman Sorry, I'm currently working with iOS app only. Could you debug and find where is the problem (probably within NavBar.js)? Maybe because of upper tabbar and absolute positioning?

quarryman:
@aksonov I will take a look and tell if i can narrow it down.

aksonov:
@quarryman Could you check latest master now?

quarryman:
@aksonov I can not confirm this has been fixed for Android on master. 
Looking into PR now

quarryman:
PR merged fixed the issue.
Still it was merged to master and Example rely on beta.xxx currently.

buncismamen:
how to use master in my project?

I've tried with yarn add https://github.com/aksonov/react-native-router-flux.git#master but it keep using the latest release

aksonov:
It is because of yarn cache. Pass latest commit ID instead of 'master'

buncismamen:
I put the commit ID 
`yarn add https://github.com/aksonov/react-native-router-flux.git#d4fff5036653e723a697f8ab28d0ce9ee387ffb8`

and got this 
```
 Warning: Installed version 4.0.0-beta.11 of 'react-native-router-flux' does not satisfy required version https://github.com/aksonov/react-native-router-flux.git#d4fff5036653e723a697f8ab28d0ce9ee387ffb8

If there is an issue running your project, please run `npm install` in /home/user/nav-playground and restart.
```

Issue #2038, opened 2017-07-15
Icon not show on Android


Clowning:
Hi, 

I have create an app with a tabbbar, on IOS everything works, icon are  print but on Android only texts are print ! 

Do you have any idea ? And do you know how can i hide text on tabs ?

Thanks

aksonov:
RNRF is just wrapper around ReactNavigation. Please check ReactNavigation first and make sure that your icons are displayed there.

Clowning:
Yes i know it, i've tried with pure ReactNavigation and it's working ! With RNRF it's not working ! , It's working on IOS but not on ANDROID.

IOS : 

<img width="422" alt="capture d ecran 2017-07-15 a 15 32 49" src="https://user-images.githubusercontent.com/2753799/28239723-e12287be-6972-11e7-80b6-e68632989f23.png">

ANDROID : 

<img width="459" alt="capture d ecran 2017-07-15 a 15 33 01" src="https://user-images.githubusercontent.com/2753799/28239726-e9e5715e-6972-11e7-8644-d8cbed163b54.png">


CODE : 

```
import React from 'react'
import { View } from 'react-native';
import { Provider } from 'react-redux';
import { createStore } from 'redux';
import Reducers from '@Redux/Reducers';
import scenes from '@scenes';
import { Actions, Scene, Router,ActionConst } from 'react-native-router-flux';
import IconMaterialCommunityIcons from 'react-native-vector-icons/MaterialCommunityIcons';
import {ORANGE} from '@Theme/colors'
import {
  Login,
  Home,
  Events,
  EventDetail,
  CreateEvent,
  Messages,
  Notifications,
  PrivateProfil,

} from '@Pages';

const tabbarConfigStyle= {
      height: 60,
      backgroundColor: '#ECF0F1',
      borderColor: '#ECF0F1',
};

const iconHome = (navigation) => (<IconMaterialCommunityIcons size={25} color={navigation.tintColor} name="map-marker" />)
const iconNotifications = (navigation) => (<IconMaterialCommunityIcons size={25} color={navigation.tintColor} name="map-marker" />)
const iconEvents = (navigation) => (<IconMaterialCommunityIcons size={25} color={navigation.tintColor} name="map-marker" />)
const iconMessages = (navigation) => (<IconMaterialCommunityIcons size={25} color={navigation.tintColor} name="map-marker" />)
[...]

PR #2043, opened 2017-07-16
Fix for #2023 - Navigation 'back' button won't render on Android

diegocouto:
In order to properly show the navbar left actions, the absolute positioning isn't used on Android devices.

This is my first contribution, so maybe I'm not following the right protocol! Please, let me know if something is wrong or can be improved. 😄

### DIFF
--- a/dist/NavBar.js
+++ b/dist/NavBar.js
@@ -37,7 +37,7 @@ null;
 return(
 _react2.default.createElement(_reactNative.TouchableOpacity,{
 testID:'backNavButton',
-style:{position:'absolute',top:0,left:0,height:50,width:70},
+style:styles.backButtonContainer,
 onPress:onPress,__source:{fileName:_jsxFileName,lineNumber:38}},
 
 _react2.default.createElement(_reactNative.View,{style:style,__source:{fileName:_jsxFileName,lineNumber:43}},
@@ -277,14 +277,14 @@ top:8}}),{
 right:2,
 paddingRight:8}),
 
-leftButton:_extends({
-position:'absolute'},
+leftButton:_extends({},
 _reactNative.Platform.select({
 ios:{
+position:'absolute',
 top:12},
 
 android:{
-top:8},
+top:0},
 
 windows:{
 top:8}}),{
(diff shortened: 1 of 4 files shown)


aksonov:
Thanks! Unfortunately circleci didn't run, I've fixed it, could you just commit something to run eslint/tests ?

aksonov:
Maybe you could look also #2038?

diegocouto:
@aksonov I can reproduce this same issue described on #2038. I'm working on it right now. 😉

diegocouto:
@aksonov I have added the backHandler to the App component on the Router, but I'm probably getting something wrong about accessing the index of the current scene.

The original:
``` javascript
const App = observer(props => {
  const AppNavigator = props.navigator;
  return (
    <AppNavigator navigation={addNavigationHelpers({ dispatch: navigationStore.dispatch, state: navigationStore.state })} />
  );
});
```

Is now defined as:
``` javascript
@observer
class App extends Component {
  componentDidMount () {
    BackHandler.addEventListener('hardwareBackPress', this.onBackPress);
  }

  componentWillUnmount () {
    BackHandler.removeEventListener('hardwareBackPress', this.onBackPress);
  }

  onBackPress = () => {
    if (navigationStore.state.index === 0) {
      return false;
    }

    navigationStore.pop();
    return true;
  }

  render () {
    const AppNavigator = this.props.navigator;
    return (
      <AppNavigator navigation={addNavigationHelpers({ dispatch: navigationStore.dispatch, state: navigationStore.state })} />
    );
  }
}
```

The issue here is that `navigationStore.state.index` is always `0`. Do you have an idea about what is going on?

Regarding #2038, I think that we would only have to update the docs. To make it work on android:

1. It's important to set `showIcon` on the tabbar root scene. Icons are disabled on Android by default, [as stated here](https://reactnavigation.org/docs/navigators/tab#tabBarOptions-for-TabBarTop-default-tab-bar-on-Android).

2. Instead on defining a scene like:
``` javascript
<Scene 
  key='home' 
  component={Home} 
  title='Home'
  iconName='ios-home'
  icon={TabIcon} hideNavBar />
```

The right way to do it now is:
```javascript
<Scene 
  key='home'  
  component={Home} 
  title='Home'
  tabBarIcon={() => (<TabIcon iconName={'ios-home'} />)} />
```

hofuchi:
How about something where if the scenes don't change: exit the app
for your `onBackPress()`
```javascript
onBackPress = () => {
  navigationStore.pop();
  if (navigationStore.currentScene === navigationStore.prevScene) {
    return false; // Exit the app if you're at the root scene
  } else {
    return true; // Keep the current popped scene
  }
}
```

diegocouto:
Thanks for all your help, @hofuchi! I've followed the approach you suggested and everything seems fine.

aksonov:
LGTM!

PR #2046, opened 2017-07-16
Migrate to CircleCI 2.0

charpeni:
Replace CircleCI configuration by CircleCI 2.0.

Steps have been optimized to use yarn and cache it between build. 

Builds should now take 20-30 seconds instead of 2 minutes.

CircleCI will fail if yarn.lock isn't updated.

cc @aksonov.


aksonov:
Speed is awesome now, thanks!

PR #2071, opened 2017-07-19
Fix children scene transition props not collected

wsxyeah:
When collecting children scene's transition props, `transition[props]` was used,
I think I should be `props[transition]`.

see code changes


aksonov:
Oh, you are right, thanks!

PR #2073, opened 2017-07-20
Add Scene Props and Etc

luco:
Add Scene props and format defaults.


aksonov:
I see duplicate `onLeft` there, could you look?

Issue #2137, opened 2017-07-29
[Feature] Deep Linking Support


kesha-antonov:
Hello!

I've seen comments here: https://github.com/aksonov/react-native-router-flux/issues/704

Can someone add deep linking in v4?

I'll try to play around. Maybe someone already has working code?

aksonov:
Okey, I added `path` passing to react-navigation, but it doesn't work good. react-navigation router doesn't return any action for deep nodes, only for top level. Also it creates several pushes for some reason... You need to debug react-navigation code to make it work. Check PR above, you may open login popup using `xcrun simctl openurl booted mychat://login/123`, but it pushes two login screens for some reason.

kesha-antonov:
Thanks!

luco:
This would be awesome.

luco:
Any updates?

shukerullah:
Any updates on this?

thg303:
I need deep linking! :(

luco:
Is there any workaround for deep link while we wait?

thg303:
I used react-router. it supports deep linking

willymulyana:
Try to use `crossroads ` follow this: [http://ihor.burlachenko.com/deep-linking-with-react-native/](url). I was try it with v4 and it works well. Hope it help.

pewh:
@willymulyana the link is broken

jasongrishkoff:
@pewh here's a working link: [http://ihor.burlachenko.com/deep-linking-with-react-native/](http://ihor.burlachenko.com/deep-linking-with-react-native/)

dittmarconsulting:
Hi @jasongrishkoff have you tried to send the URL via email or SMS and close the app (iOS swipe up) What happens when you click the link? Will it just open the app or open the app and go to the "Hello World" scene?

kesha-antonov:
https://github.com/aksonov/react-native-router-flux/pull/2148#issuecomment-342881209

Thanks! @aksonov

PR #2148, opened 2017-07-31
Deep linking

aksonov:
#2137


aksonov:
It seems to work (Example app):
`xcrun simctl openurl booted "mychat://root/register"`

Issue #2165, opened 2017-08-02
Tab bar not showing in Android 


nakulkundaliya:
Hello @aksonov 

I am facing issue in android. Tabbar doesn't display in android device.  I have checked your demo app code. It's working fine in IOS but, I facing issue in android.

I attached two screenshot 1st is android and 2nd is IOS/

Please help me.

![screenshot at aug 02 14-50-17](https://user-images.githubusercontent.com/15937642/28867118-a695c59c-7792-11e7-8535-be949f5fb7ea.png)
![screenshot at aug 02 14-58-22](https://user-images.githubusercontent.com/15937642/28867269-11b511de-7793-11e7-9a33-59c97803166a.png)

mikaello:
I am experiencing the same thing. Samsung S7, Android 7.0.

aksonov:
Try to run pure react-navigation tab bar demo and check. RNRF just uses react-navigation TabNavigator internally.

aksonov:
https://github.com/react-community/react-navigation/issues/1420

nakulkundaliya:
Thanks for link
But i used **'react-native-router-flux'** from page nevigation. let me know how it's possible to use  '**react-navigation**'

aksonov:
RNRF uses react-navigation internally.

nakulkundaliya:
Okay,
So how Can i fixed it. you have any idea about this?

luco:
It seems that it's something related to styles. In fact, the tabbar appears, but it's not properly styled, as you can see below:

![image](https://user-images.githubusercontent.com/246373/28904750-813b9d2a-77e3-11e7-92c2-9765d3a05639.png)

aksonov:
@nakulkundaliya First I need to know where is the bug - in RNRF or within react-navigation. That is why I asked you to try react-navigation tabbar demos first. If the bug is there, create issue there please.

nakulkundaliya:
@aksonov  i agree @luco tab is appear but not display properly. 
How can i set tabbar at bottom ( Android)?

One more thing if i set `showLabel={true}` then tab is appear with label but its display at top of the screen.

aksonov:
Then probably you have to set proper styling. Definitely react-navigation issue...

nakulkundaliya:
Can you please suggest to me how can i fixed it?

Here is my router code
```
<Scene
  key="tabbar"
  gestureEnabled={false}
  showLabel={false}
  tabs
[...]

PR #2168, opened 2017-08-02
Update swipeEnabled and Etc.

luco:


aksonov:
Thank you for a PR! It seems there options are not only for Android, could you remove 'Android' word?

luco:
Sure. I thought it was Android only because I only saw when tested on Android.

aksonov:
I don't have Android sdk, could you please check #2165 ? Does tabbar really doesn't show on Android?

PR #2179, opened 2017-08-03
Fixed popTo method

ulentini:
`popTo()` method used to fail in case there were a Scene in the stack of the same kind of the current one.

**How to reproduce it**
Let's assume we have a navigation stack with the following Scenes:
`[A, B, C, C]` (being A the root Scene, C the current Scene).
Calling `popTo('A')` or `popTo('B')` would fail, because of `nextScene !== currentScene` within the `while` condition. Removing that condition solved the problem.


afilp:
@ulentini I have the following situation:

[A, B, B, B, C, D]

"B" is actually may "app controller" which does different things based on different props.

I want to pop to the 1st "B" state.

How can I use the `popTo` for this?

Thanks!

Issue #2195, opened 2017-08-07
Access to Component's props from NavBar.


tendolukwago:
### Version
Tell us which versions you are using: 

- react-native-router-flux v4.0.0-beta.15 (v3 is not supported)
- react-native v0.46.4?

### Expected behaviour

First, great library. it's been a ton of help so far. There is some important functionality though, that I'd like some clarification on. 

In v4, I expect to be able to call a static method that has access to the component's props. 

Being able to save and interact with the store, seems like pretty necessary functionality....maybe that's just me, though. I've seen you discuss this in a couple of threads, but there hasn't been any clear solution. If it's the custom navigation that's the answer, can you please provide an example of how to use it with the rnrf? The documentation isn't very clear. 

Thanks.

### Actual behaviour
I can't access the component's props.

aksonov:
Check 4.0.0-beta.18 - now you can access your component instance via `Actions.refs.SCENE_NAME`

PR #2248, opened 2017-08-16
access to screen components via `Actions.refs`

aksonov:
access to screen components via `Actions.refs` (#2195)

### DIFF
--- a/src/navigationStore.js
+++ b/src/navigationStore.js
@@ -36,6 +36,9 @@ export const actionMap = {
 
 const reservedKeys = [
   'children',
+  'refs',
+  'addRef',
+  'removeRef',
   'create',
   'execute',
   'popTo',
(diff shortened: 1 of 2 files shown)


src/navigationStore.js
```diff
@@ -190,19 +193,40 @@ function createNavigationOptions(params) {
...
+      }
+    }
+    componentWillUnmount() {
+      const navigation = this.props.navigation;
+      this.ref = null;
+      store.deleteRef(originalRouteName(navigation.state.routeName));
+    }
+    render() {
+      const navigation = this.props.navigation;
+      return <Component ref={ref => (this.ref = ref)} {...this.props} {...navigation.state.params} name={navigation.state.routeName} />;
```
davojan: @aksonov This causes React warning if the Component is stateless:
```
Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.
```

aksonov: Good point. However making Screens as pure function is not good idea - hot reload will not work for them...

PR #2258, opened 2017-08-17
Prevent React warning (and potential errors) about refs for stateless function components

davojan:
Disabled screen refs support for stateless function components because react doesn't support refs for them.

Without this fix there is a warning:
```
Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.
```


aksonov:
Looks great!

Issue #2284, opened 2017-08-24
No hard dependencies on mobx


RWOverdijk:
### Regarding V4

Currently, MobX is a hard dependency on the library. For those not using MobX (me being one of those people) that's not something that's wanted.

### Proposal

It's a small proposal, because I haven't worked out the specifics yet. But basically:

* Remove MobX as a dependency
* Document "engines" to use
* User is responsible for including dependency in the project
* User tells RNRF what engine to use (for the specific methods)

This could look something like:

```jsx
import {ENGINE_MOBX} from 'react-native-router-flux'

<Router engine={ENGINE_MOBX} />
```

This would tag in MobX behavior here:

- https://github.com/aksonov/react-native-router-flux/blob/master/src/navigationStore.js
- https://github.com/aksonov/react-native-router-flux/blob/master/src/Router.js

Thanks to @rmevans9 and @skellock for sharing this opinion and figuring out where MobX lives in the project.

skellock:
I'm team-`mobx` and I still want this for my `redux` compatriots.  Lemme know if I can help at all.

aksonov:
Okey, any idea how to implement Router/navigationStore without mobx ?

rmevans9:
Hey @aksonov!

We are still working out those details but after looking through the code it doesn't seem like an unattainable goal. @skellock @RWOverdijk and I will be working through a few of the options we have talked about in the next few days.

rmevans9:
Hey @aksonov - I have gotten some work done on playing around with the proposal that @RWOverdijk suggested. I haven't opened a PR on this repo yet as there is a looottt of cleanup to do in addition to getting a good example of the `ReduxEngine` in place. I created a PR on my fork of this repo so @skellock and @RWOverdijk could see/contribute to our proposed solution. In the mean time please feel free to take a look at the direction it is heading:

https://github.com/rmevans9/react-native-router-flux/pull/1

The current state of it is - it works! I was easily able to pull `MobX` out of `Router` and `navigationStore` while still providing access back into them so navigation still works. I also built out a `ReduxEngine` that also works but requires a little bit more setup then the `MobX` one. I personally think I can improve the API that it is currently using to make the setup a little bit better on the `redux` side.

[...]

PR #2293, opened 2017-08-28
Make the state engine configurable so there is no hard dependency on MobX

rmevans9:
This is the first pass at pulling `MobX` from being a hard dep in `react-native-router-flux` and instead become an optional one and allow for `redux` to take its place. Cursing around the example everything appears to work correctly with these changes. Note: I did some cleanup in the example app to allow for it to showcase both `MobX` and `redux` state setups.

Fixes: #2284


rmevans9:
I will work on fixing tests in the next day or two.

aksonov:
I was thinking about removing mobx/redux dependency at all... What about passing optional 'setState' function to navigationStore (default React.Component.setState will be used)?

aksonov:
Ignore my last comment about setState - it will not work for navbar buttons. I still don't want to increase complexity for 90% of users (who doesn't care about internal mobx), could you found a way to   have both worlds - default (mobx) and customised (redux or anything else). I think it could be achieved with `require` usage (not import), what do you think?

rmevans9:
I will give it another shot tonight and see if I can gain any traction.

rmevans9:
Hey @aksonov - sorry it took a bit to get back to this. It turns out that the react native packager doesn't want to optimize `mobx` out at all if you have it in as a dependency of `react-native-router-flux`. So as I see it the options are:

1. Do nothing - let `mobx` be a dep and if someone wants to use this lib they have to accept that
2. Make this breaking change which should remove the dep, if it doesn't you might need to pull the packages out to their own packages that depend on a `core` library.
3. Figure out the best plan to remove all external state tree implementations.

I feel like the 3rd option would be the absolute best plan moving forward from here but would require the most work likely.

aksonov:
Okey, could you add default state engine based on `Component.setState`? This way we could introduce
breaking change for mobs-users who use `wrapBy` (i.e. require to pass `engine` instead of current `wrapBy`)

rmevans9:
I can work on that in the coming week

rmevans9:
I have updated this PR with the first pass at providing a "PureEngine" which is just regular 'ol state. Because of the way the `navigationStore` is setup there is some working around that happens but I think that can be worked out to be better in time.

Things I still want to do:
1. Test this more.
2. Figure out the best way to ship the other two engines in separate packages. (Different Repos that depend on this one? Setup for them to be shipped as individual packages?) - Guidance from @aksonov would be appreciated here since ultimately it will be him that has to manage this.
3. Review the code I wrote this morning a bit more to see if I can do things better and not in such a round about way.

aksonov:
@rmevans9 Thanks! Looks interesting, code seems to be ok, unfortunately I was quite busy these days and will have vacation starting Sep, 20. Could you resolve conflicts? Also could you check how many times components are mount with redux/standard engine - issue #2382 ? Could it happen because of mobx?

rmevans9:
Merged in master to resolve conflicts. I put it on my todo list to test out #2382 but won't have time till tonight or possibly tomorrow.

rmevans9:
@aksonov The Drawerer still gets mounted multiple times even with the "PureEngine". I will spend a little bit of time digging into why that might be.

rmevans9:
Scratch that - I was seeing the issue because of the fact that in the example `TabView` is actually used for basically all the tabs so it really was mounting multiple for each tab. I do not see it mounting multiple times incorrectly once I switched the wrapper to its own component.

rmevans9:
@aksonov I am actually going to close this PR. Although it was a direction to resolve the mobx dependency I actually found a better why to implement it. I will be opening a new PR in the next week that will remove any mobx code from RNRF and let `react-navigation` handle the state internally unless a user provides their own state and dispatch functions. The change will likely effect anyone passing in a state and dispatch function but should be something that is transparent to users that are not.

Issue #2261, opened 2017-08-18
There is no route defined for key {route}.


signorekai:
### Version
Tell us which versions you are using: 

- react-native-router-flux v4.0.0-beta.18
- react-native v0.45.1

### Expected behaviour

Actions.{route}() should work and navigate me to scene.

### Actual behaviour

Throws error ```ExceptionsManager.js:63 There is no route defined for key {route}.
Must be one of: 'root','newBroadcast'```

### Steps to reproduce
For non-obvious bugs, please fork this component, modify Example project to reproduce your issue and include link here.

This is my <Router> routes.

```
<Router>
  <Scene modal={true} hideNavBar>
    <Stack key='root' renderTitle={navBarLogo}>
    <Scene key='loading' component={LoadingPage} hideNavBar={true} inital />
    <Scene key='splash' component={NewRegisterPage} type="replace" hideNavBar={true}  />
    <Scene key='feed' component={FeedPage} type="reset"  />
  </Stack>
  <Scene key='newBroadcast' component={NewBroadcastPage} direction='vertical' />
  </Scene>
</Router>
```

`Actions.feed()` throw me the no route defined error. I'm at my wit's end and would appreciate any form of help.

signorekai:
After some diagnosing, I realise the `type='reset'` is the one causing the issue. Does anyone have any idea on how to fix this?

aksonov:
Have you tried to use `replace` instead?

signorekai:
Hey @aksonov, not sure why I haven't tried this previously but using `type={ActionConst.REPLACE}` fixed the issue. Will submit a PR for the docs

ZeroCool00:
anyone solve this error..  replace is not working.

daviscabral:
Please open a new issue with the details of your error and sample code to help someone to jump on this.

ZeroCool00:
#3104

ZeroCool00:
@daviscabral i open this issue here #3104

PR #2344, opened 2017-09-05
Fix for reset actions being unable to find nested scenes

seratonik:
Found that when wrapping scenes in a Modal, Lightbox or Overlay that _only_ the top level keys for those wrappers were available to perform a `reset` action to. The `replace` action wasn't sufficient as it still allowed swiping from the left to navigate back to the previous scene.

This PR fixes this issue and allows a `reset` to any key just like `push`, etc do.

However, I'm uncertain if this potentially breaks any other functionality - and it should definitely be reviewed thoroughly. Potentially fixes closed issue #2261 that was worked around.

### DIFF
--- a/dist/navigationStore.js
+++ b/dist/navigationStore.js
@@ -600,7 +600,7 @@ _this3.dispatch({type:ActionConst.REPLACE,routeName:routeName,params:params});
 
 reset=function(routeName,data){
 var params=filterParam(data);
-_this3.dispatch(_reactNavigation.NavigationActions.reset({key:null,index:0,actions:[_reactNavigation.NavigationActions.navigate({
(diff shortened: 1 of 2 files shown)


mcabs3:
@seratonik although this fixed resetting nested scenes, however this breaks resetting from a nested scene to another scene.

seratonik:
@mcabs3 shoot, I haven't experienced that yet in my project. I use the reset upon login, and cold-boots to send people to a tabbed area asynchronously, and then once again when they logout to go back to a guest experience and it seems okay right now.

aksonov:
Yes, 'react-navigation' reset action is very weird. Sorry I cannot merge because we had this change before. 

The correct way would be to write own processing within reducer to fit both cases (reseting from nested scene to another scene AND @seratonik ) ...

I have quite limited time now to do it, but it is already implemented for v3 version - you just need to convert it and put that logic to v4 Reducer

PR #2356, opened 2017-09-08
Fix bug that replace with tab'ed Scene works like push

Pylipala:
Trace up toward parent from active scene, until `popPrevious` could be done.

### DIFF
--- a/dist/State.js
+++ b/dist/State.js
@@ -43,9 +43,12 @@ inject=inject;exports.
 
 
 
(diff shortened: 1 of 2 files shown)


mcabs3:
@Pylipala looks like your PR failed eslint formatting. Please correct so that the CircleCI tests pass

mcabs3:
@aksonov I like this change though, seems to work from my tests.

mcabs3:
The change has already been merged into master

PR #2364, opened 2017-09-10
Updated API Documentation & Added fix for replace Tabbed Component

mcabs3:


mcabs3:
@aksonov I updated the docs from the notes from previous issues, and I also added a separate commit for the replace issue for `Tab` components since the other PR wasn't passing the CircleCI tests.

PR #2390, opened 2017-09-15
Ensure example works by referencing specific version of react-navigation

arcseldon:
Discovered that the example would error due to conflicts with React Navigation. Chose to specify the exact version in the example package.json (as originally supplied by the example author), and the sample ran without any issues. Hope this helps others until ready to upgrade example with all latest changes to move to most recent versions.


mcabs3:
Thank you for taking the time to make a PR. We just merged another commit to resolve this.

PR #2398, opened 2017-09-18
Allow to hide left button via Actions.refresh

onrige:
Hide left button on the fly via Actions.refresh by `left`, `leftButton` and `renderLeftButton` properties.

### DIFF
--- a/src/navigationStore.js
+++ b/src/navigationStore.js
@@ -188,12 +188,15 @@ function createNavigationOptions(params) {
         || (init ? null : (renderBackButton && renderBackButton(state)) || <BackNavBarButton {...state} />);
     }
 
-    if (typeof navigationParams.back !== 'undefined') {
-      if (!navigationParams.back) {
+    if (back) {
+      res.headerLeft = (renderBackButton && renderBackButton(state)) || <BackNavBarButton {...state} />;
+    }
+
+    if (typeof navigationParams.left !== 'undefined' || typeof navigationParams.leftButton !== 'undefined' ||
+      typeof navigationParams.renderLeftButton !== 'undefined') {
+      if (navigationParams.left === null || navigationParams.leftButton === null || navigationParams.renderLeftButton === null) {
         res.headerLeft = null;
       }
-    } else if (back) {
-      res.headerLeft = (renderBackButton && renderBackButton(state)) || <BackNavBarButton {...state} />;
     }
 
     // currect dynamic navigation params has priority over static scene params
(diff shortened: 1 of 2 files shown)


src/navigationStore.js
```diff
@@ -188,7 +188,11 @@ function createNavigationOptions(params) {
         || (init ? null : (renderBackButton && renderBackButton(state)) || <BackNavBarButton {...state} />);
     }
 
-    if (back) {
+    if (navigationParams.back != null) {
```
mcabs3: please follow JS standards and use `!==` instead of `!=` as well as `!navigationParams.back` instead of `=== false`

onrige: Right, thx. But first one was a typo by the way.

src/navigationStore.js
```diff
@@ -188,7 +188,11 @@ function createNavigationOptions(params) {
         || (init ? null : (renderBackButton && renderBackButton(state)) || <BackNavBarButton {...state} />);
     }
 
-    if (back) {
+    if (typeof navigationParams.back !== 'undefined') {
+      if (!navigationParams.back) {
+        res.headerLeft = null;
```
davojan: This seems odd to me, back == false should not hide left button according to API docs. 

Proper support for changing back param via Action.refresh would look like this:

```js
if (navigationParams.back || (navigationParams.back !== false && back)) {
  res.headerLeft = (renderBackButton && renderBackButton(state)) || <BackNavBarButton {...state} />;
}
```

But this will not hide back button in all cases I believe.

onrige: @davojan It seems like Docs where changed since I did this PR. So I really need this functionality (and may be not only me in future), may be you have some suggestions about this? This add is doing the job i want for now.

davojan: Not sure but I think with my code above ``Action.refresh({back: false})`` should work for you if you haven't defined any of those left-related options for the scene (``leftButton``, ``onLeft``, ``leftTitle`` an so on, see https://github.com/onrige/react-native-router-flux/blob/226f3b645da2b9ef298e57eb2e8296f817880a50/src/navigationStore.js#L182).

If you want to hide left-button (not back-button) I think the right way is ``left=false`` or something like that, but ``back=false`` doesn't make sense in this case.

onrige: You're right. In my case it's won't work because I'm using leftButton. I think I'll update this PR with hide `left` functionality instead of `back` in the evening. Thanks.

PR #2407, opened 2017-09-20
Fix crash when unmounting WrappedComponent

wsxyeah:
Check if has routeName to prevent crash

### DIFF
--- a/dist/navigationStore.js
+++ b/dist/navigationStore.js
@@ -213,7 +213,7 @@ return res;
 };
 }
 function originalRouteName(routeName){
-if(routeName.startsWith('_')){
+if(typeof routeName==='string'&&routeName.startsWith('_')){
 return routeName.substring(1);
 }
 return routeName;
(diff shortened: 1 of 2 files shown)


mcabs3:
Do you have an example of what would cause this crash?

wsxyeah:
@mcabs3 Yes.

Add a `key` for the container in Example.js (just like the `LightBox`).
Then choose the RN menu `Toggle Inspector` on the device, you will see the crash.

```javascript
<Router>
  <Modal>
    <LightBox key="lightbox_top"> // here
      <Stack key="root"/>
      // other scenes …
    </LightBox>
  </Modal>
</Router>
```

wsxyeah:
@mcabs3 Oh, I found the root cause now.

Sine the `OverlayNavigator` passed wrong `state` to its child overlay components, the children will never get `this.props.navigation.state.routeName`.

mcabs3:
@wsxyeah so is this change irrelevant now?

PR #2422, opened 2017-09-24
Fix passing down wrong overlay navigation state in OverlayNavigator

wsxyeah:
We shouldn't pass the down the `state` directly to the overlay children,
instead, pass the corresponding state in `state.routes`.

This bug also caused the issue mentioned  in  #2407.

### DIFF
--- a/dist/OverlayNavigator.js
+++ b/dist/OverlayNavigator.js
@@ -22,11 +22,13 @@ routes=state.routes;
 
 var Component=routeConfigs[tabsConfig.initialRouteName].screen;
 var initialIndex=0;
+var routesMap={};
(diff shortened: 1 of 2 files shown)


mcabs3:
`
/home/circleci/react-native-router-flux/src/OverlayNavigator.js
  25:26  error  Trailing spaces not allowed  no-trailing-spaces
  31:17  error  'routeName' is not defined   no-undef
  31:35  error  Trailing spaces not allowed  no-trailing-spaces
  31:35  error  Missing semicolon            semi
`

wsxyeah:
@mcabs3 fixed

PR #2426, opened 2017-09-24
Remove MobX dependency

rmevans9:
This is another take at my original PR instead of introducing a new concept (The "Engines") this reworks the logic to hook into `react-navigation` in a different way which allows it to work without passing a custom navigation object which means by default the state is managed by `react-navigation`. 

There are a few things that I still need to complete or test:
* Finish todos left in code
* Test if someone does pass in a navigation implementation (like MobX or Redux)
* Further testing
* It seems one of the examples fails - the "Back and Refresh" from Login2. Figure out if we can make that work

A few thoughts after combing through the codebase:
* Since `react-navigation` is back to being maintained it might make sense to evaluate some of the logic in this and see if it would make sense to move it upstream. I will compile a list of what would make sense for this in the coming days.
* This refactor to get rid of MobX seems small but likely will have a fairly big impact. It should be throughly tested.

## Edit
A running list of things that still need to be done
- [x] Fix issue with initial screen getting onFocus event hook
- [x] Re-implement deep linking logic
- [x] Either remove or rewrite test - NOTE: This is actually broken because of the last item in this list. The test might go green again once that issue is fixed
- [x] Test providing a state implementation (similar to how `react-navigation` allows you)
- [x] Find and finish any todos I left in the codebase
- [x] Post merge the `pop` action doesn't seem to work right
- [x] Re-implement "no double pushes" logic (most likely needs to move to Reducer)
- [x] Evaluate ways to not use the internal `_navigation.state` in `Router.js`


christianbach:
any updates?

aksonov:
@rmevans9 Just returned from vacation. Have you tried to solve "Back and Refresh"? Any other problems?

rmevans9:
@aksonov I haven't had much time to work on the issues with this PR. I hope to in the next week or two.

rmevans9:
I have updated this to master, found out why "Back and Refresh" was not working and put in a fix for that (a kinda hacky one). I have not been able to get a MobX or Redux implementation going with this new setup but that is solely because I haven't spent any time on it yet. I am not sure how much more time I will be able to dedicate to this PR so if someone wants to pick up the torch by all means.

aksonov:
Looks great, Example works well, but my app doesn't work  - there is something wrong with onEnter/success/failure transitions. Probably setTimeout will help, will try to fix that...

rmevans9:
@kesha-antonov One of the biggest reasons I did this work is to just remove the hard dependency on it for every app that uses Redux. If you use `react-native-router-flux` you now have a `mobx` dependency even if you didn't want it.

rmevans9:
@aksonov if you can see if you can get the failure you are seeing happening in the Example I am more then happy to take a look at that as I think it should be a simple fix. Probably just a timing issue.

aksonov:
@kesha-antonov Could you check how it works on your app?

aksonov:
@rmevans9 Just replace 'launch' scene with 
`            <Scene key="launch" component={Launch} title="Launch" initial on={() => console.log("ONENTER!")} success='register' failure='customNavBar' />`
and you will see that onEnter/on is not executed at all.

kesha-antonov:
@aksonov Sure. I will check tomorrow

rmevans9:
@aksonov Got it. I will see what is going on there.

rmevans9:
@aksonov the issue seems to be related to the fact that the initial screen doesn't have the `onNavigationStateChange` called by `react-navigation`. I set it on the Register screen and it appeared to work and even having it on the initial screen then navigating to a different screen then going back triggers the event correctly.

All that should need to happen is to figure out how to trigger a onNavigationStateChange upon initial load of the app I believe.

aksonov:
@rmevans9 Yes, I found the same - onNavigationStateChange is not called initially. Probably we could call it from navigationStore.create, what do you think?

P.S. Omg, got conflicts again now after deep-linking PR and updating Example to latest RN.. Maybe we could remove `dist` generated files at all, because auto-merge doesn't work with generated files...

rmevans9:
I've pushed up the merge but I basically threw out the deep link things because it would likely require some manual intervention. So we will need to re-implement that in the new way that Router is setup. Shouldn't be hard to pull off.

I think dumping the dist folder from git makes a lot of sense. I can go ahead and do that in this PR if you like.

rmevans9:
Looks like there is also a test failing post merge. I will sort that out tonight.

rmevans9:
Ah - the test is failing because it is actually incompatible with the way the state is being tracked now. The tests tries to look at it directly but with the new setup lets `react-navigation` handle the store. So the test needs to be rewritten most likely.

rmevans9:
Updated the original comment with a running list of things that I think are still left for this PR.

rmevans9:
Got the pop command working again which fixed the test

rmevans9:
Initial screen getting the `FOCUS` and the `onEnter` event getting called now

rmevans9:
I have added deep linking back based on the PR but have not tested.

aksonov:
Looks great! One more thing is to change package.json to point to src/ ...

aksonov:
Oh, you removed Actions.currentScene and Actions.prevScene as well as double actions preventing.
Another issue that you are using `navigator._navigation.state` that is hidden property and could be changed by react-navigation without any notice. Why don't use own state of `Router` component for navigation state as well as currentScene/prevScene (i.e. update all them at once with `Router.setState` method?

rmevans9:
I can certainly update the package.json to point to src.

I did remove currentScene, prevScene and storing the state locally since the point of this change was to offload the state management to `react-navigation`. We can keep state locally and try and keep them in sync between the two. The reason I avoided that was because whenever I have tried to keep two sets of state in sync in other cases you always run into edge cases that could get them out of sync. Thinking through the use case here I think it could work though. That would remove the need for using the hidden `navigator._navigation.state` use as well.

I am doing some work to make sure this works with Redux and going to try and do MobX next. After I wrap that up I will see about adjusting the way we handle the state here.

rmevans9:
Got it working with a Redux state setup! I haven't played with getting MobX to work yet but I suspect it should be pretty easy now that I got the kinks worked out to get it working with Redux.

rmevans9:
@aksonov The last few changes left are:
- Don't use the `_navigation.state` internals.
- Re-add Actions.currentScene and Actions.prevScene

For avoiding the use of `_navigation.state` we could tap into the state of the AppNavigator (which ends up being whatever the root level navigator is). I personally like that more because 1. we are no longer using some internals of `react-navigation` and 2. We are not duplicating the state in two places.

For `Actions.currentScene` and `Actions.prevScene` we can track that easily but it won't be MobX observable or anything like that. Does that work?

rmevans9:
I've went ahead and implemented the prevScene and currentScene. I have also switched how I am using the state from `react-navigation`.

src/Router.js
```diff
@@ -44,7 +44,7 @@ export default class Router extends React.Component {
   componentDidMount() {
     if (!this.props.dispatch) {
       navigationStore.dispatch = this.navigator._navigation.dispatch;
```
aksonov: Could we avoid using private properties like `_navigation` ?

src/Router.js
```diff
@@ -44,7 +44,7 @@ export default class Router extends React.Component {
...
-      navigationStore.getState = () => this.navigator._navigation.state;
+      navigationStore.getState = () => this.navigator.state.nav;
```
aksonov: I still don't understand why we can't use own state? This way react-navigation should not maintain own one (as their document says).

aksonov: The problem also that main idea of mobx integration was to make `currentScene` observable, i.e. re-render screens that use `currentScene`. It is usable for example to display `loading` buttons if `prevScene != currentScene`. I will check how it works with mobx-less implementation...

src/Router.js
rmevans9: Ah I missed this one. I will see if I can get this one out of here too.

src/Router.js
rmevans9: I am trying to avoid having our own state because it brings little benefit to try and maintain it ourselves vs letting `react-navigation` handle it. 

Without `MobX` or something of that nature I don't think you will be able to easily cause a re-render when `prevScene` or `currentScene` change.

daviscabral:
It has been a while since this PR - but I would like to check about the state of this since we are moving into something similar with the migration to `react-navigation@2`.

aksonov:
It is different - own state management and it is not compatible with v2...

aksonov:
Finally 4.0.0-beta.40 is released (based on react navigation 2.x), without mobx dependency.

PR #2491, opened 2017-10-10
Prevent displaying of BackButton in iOS style on Android when icon is not customized

davojan:
When rnrf's BackButton was used instead of the react-navigation's one on Android back-button was displayed in iOS style which was ugly.

I made rnrf ``BackButton`` to fallback to rn's ``HeaderBackButton`` component if custom ``backButtonImage`` is not provided. If I understood right own ``BackButton`` was introduced only for that purpose.

In my opinion better way is to rewrite ``BackButton`` completely by copying ``HeaderBackButton`` from ``react-navigation`` and adding customisation features but keeping styles. The styles are done pretty good in ``react-navigation``. But this requires more time investment and it could break something that I'm not aware of.

### DIFF
--- a/dist/NavBar.js
+++ b/dist/NavBar.js
@@ -12,7 +12,22 @@ Object.defineProperty(exports,"__esModule",{value:true});var _extends=Object.ass
 
 
 
+
 BackButton=BackButton;exports.
+
+
+
+
+
+
+
+
+
+
+
+
+
+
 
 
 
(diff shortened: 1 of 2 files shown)


aksonov:
Looks like better fix just to submit android images for back button image (if there is only difference in default images)?

davojan:
Nope. Also whole bunch of custom styling is involved and backTitle should never be shown on android.

Can you tell what was the intention of introducing the custom ``BackButton``. Only ability to customise the icon or something else?

aksonov:
Probably only because of icon.

Issue #2441, opened 2017-09-27
How to hide navBar pragmatically


pgonzalez-santiago:
### Version
- react-native-router-flux v4.0.0-beta.21 (v3 is not supported)
- react-native v0.48.3,

I need to hide the navBar inside the component. I have tried  to run: 

`Actions.refresh({ hideNavBar: true })`

but it doesn't work. Any ideas of how can I hide the navBar programmatically?

MichaelPintos:
`<Scene key="myView" component={myView} hideNavBar={true}/>`

in myView you will need to create your own navigation bar.

```
<View style={styles.container}>
        {this.state.renderNav && (
          <View>
            <Text>NavBar</Text>
          </View>
        )}
        <Text>MyView</Text>
        <TouchableHighlight onPress={() => this.setState({ renderNav: !this.state.renderNav })}>
          <Text>Toggle Nav</Text>
        </TouchableHighlight>
 </View>
```

onrige:
@pgonzalez-santiago This feature was able about two weeks ago by my PR but it's broke drawer in some cases it seems.
So you can have a look at commit: https://github.com/aksonov/react-native-router-flux/commit/29090461469d4eb907072c63e9881f99205fab73

pgonzalez-santiago:
So what do we do now? I am doing feature that needs fullscreen so I have to hide the navbar.

onrige:
@pgonzalez-santiago Fine, I will try to reproduce this bug in commit and resolve it. But you must admit that there is no accepted PRs since @aksonov went on vacation (till 8 october).

ihor:
I'm having the same issue but I want to display navBar on scroll.

thiemo--bleeken:
have the same issue. how can we hide the navbar?

alexhoo:
Same issue here :(

jaywalklabs:
```
<Router 
navBar={NavBar} >
....
</Router>

export default class CustomNavBar extends Component {
    render() {
        const { title, hideNavBar } = this.props;
        if(hideNavBar === true) return <View />
        return (
            <View>
                <Text>{title}</Text>
            </View>
        );
    }
}
```
then whenever you can set

`Actions.refresh({hideNavBar:true})`

onrige:
@pgonzalez-santiago @ihor @thiemo--bleeken @alexhoo now you can monitor any updates on this here #2492.

pgonzalez-santiago:
Thank you @onrige

[...]

PR #2492, opened 2017-10-10
Hide navbar programmatically

onrige:
According to this issue #2441 and this commit (https://github.com/aksonov/react-native-router-flux/commit/29090461469d4eb907072c63e9881f99205fab73) I am adding this functionallity again. I was trying to catch the issue when NavBar was hiding inside Drawer but I didn't succeed. May be I don't know something but according to the React Navigation docs it's not possible to put NavBar inside Drawer. @aksonov could you provide more detailed example for me to catch this error? As I said in commit comment - there are a lot of people that need this functionality so we need to resolve this problem.

### DIFF
--- a/Example/Example.js
+++ b/Example/Example.js
@@ -130,7 +130,7 @@ const Example = () => (
                 Wrapper Scene needed to fix a bug where the tabs would
                 reload as a modal ontop of itself
               */}
-              <Scene hideNavBar>
+              <Scene>
                 <Tabs
                   key="tabbar"
                   swipeEnabled
@@ -155,7 +155,6 @@ const Example = () => (
                       title="Tab #1_1"
                       onRight={() => alert('Right button')}
                       rightTitle="Right"
-
                     />
 
                     <Scene
--- a/Example/components/TabView.js
+++ b/Example/components/TabView.js
@@ -26,6 +26,14 @@ const styles = StyleSheet.create({
 });
 
 class TabView extends React.Component {
+  state = { hideNavBar: false }
+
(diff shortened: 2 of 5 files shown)


thiemo--bleeken:
please merge

aksonov:
Can you just check Example project with your PR? You will see no drawer button there, sorry, I can't break this.

just replace
```
    "react-native-router-flux": "file:../",
```
to
```
    "react-native-router-flux": "github:aksonov/react-native-router-flux#1364b111cff4fa218c9521e13a780e1228a6a708",
```
within `Example/package.json` to reproduce

onrige:
Fine. So please have a look at the last update. There was a problem with hiding `Drawer` because of `hideNavBar` property (child inheritance). It was fixed by just removing this property from `Drawer` child (`Scene` component in case of Example project) and now all works good. I am also added an example to TabView component with toggle NavBar functionality.

aksonov:
Yes, it still has problems - just tap "next screen for tab2_1" and you will not see "back" nav button.

aksonov:
@onrige Please check latest master, I hope I fixed that issue and your code to Example/navigationStore. Feel free to reopen if you have any bugs.

SurpassRabbit:
If I want to hide hideNavBar, the first success, the second failed.

```
_editPress = () => {
this.setState({
            hideTabBar: !this.state.hideTabBar
            }, () => {
                Actions.refresh({
                    hideTabBar: this.state.hideTabBar
                });
            }
        );
}
```

PR #2530, opened 2017-10-22
Whitelist event listeners to prevent "scene 'on...' is not defined" warning

kesha-antonov:
Hello!

When I open a scene with ```Actions.scene({ onUpdateSmth: this.myFunc })``` or ```Actions.scene({ onUpdateSmth: undefined })``` I get ```scene "onUpdateSmth" is not defined```.

We should use whitelist for events that starts with ```on```.

### DIFF
--- a/dist/Router.js
+++ b/dist/Router.js
@@ -59,7 +59,7 @@ scenes:_propTypes2.default.func,
 navigator:_propTypes2.default.func,
 wrapBy:_propTypes2.default.func,
 getSceneStyle:_propTypes2.default.func,
-sceneStyle:_propTypes2.default.object,
+sceneStyle:_reactNative.ViewPropTypes.style,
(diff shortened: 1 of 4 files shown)


onrige:
Hey. Could you remove package-lock from this PR, please?

kesha-antonov:
Why? It's not in gitignore and it's good practice to keep .lock files.
@onrige

onrige:
Sure, but we use yarn (also for this). You can add package-lock to gitignore in commit if you want :)

aksonov:
Please check latest master now. I fixed it in different way because you PR would break other functionality (like custom transition shortcuts like onLogin='loginScene')

kesha-antonov:
Ok. Thanks! @aksonov 
Works without PR

Issue #69, opened 2015-12-14
Contributors wanted!


aksonov:
I want to make my and your development easier with this component, but i can't spend much time on it. I do my best to make the component better and more robust, thanks for everybody for created issues, it is very valuable!

Unfortunately, number of PRs is minimal, even more, some guys just silently clone the component (even not fork), add some improvements and start promoting it as own one even without preserving copyright (however thanks for mention me as person who "inspire" them ;). I believe it is wrong way for open source development, we need to unite all our efforts to create something really great.

If somebody wants to do many changes (fix issues, write docs/unit tests) without waiting me to accept it, i'm ready to add him as collaborator.

MITDelian:
I'd be interested! I'm not sure how much time I can dedicate as well but since we're using this library in our production app I will probably be motivated to fix issues that arise

chetstone:
I was thinking about trying to help with documentation. There's still an awful lot I don't understand about this component, but perhaps I could take a stab at rewriting the readme and you could review it.

aksonov:
@MITDelian you are welcome (added)

aksonov:
@chetstone You are welcome, help with documentation is really necessary. Btw, I've added you as collaborator as well.

MITDelian:
@aksonov I thought it might be helpful to have a chat room. I've seen gitter.im used often so I set one up. just might chat some things as I'm learning the component: https://gitter.im/aksonov/react-native-router-flux

chetstone:
@MITDelian Thanks for the chatroom. See you there.
@aksonov Thanks. But I'll want you to review docs carefully please before pushing.

MITDelian:
@chetstone happy to help review as well. I'll spend some time this weekend really trying to understand the component. One of my major goals is to make integration with a pre-existing redux store much easier

markburns:
Happy to help out. Just picking up a project again that uses this, and found some incompatibility with the latest react-native. Thought I'd browse the issues to see if it's reported and found this issue. Even better if I can just fix it as soon as I understand the problem.

thachp:
@aksonov - I will be happy to help out.

aksonov:
@thachp @markburns thanks for help. Please check latest issues and try to create PR for them.

smallsaucepan:
Hi @aksonov. I wouldn't mind taking a look at #62 if you still want some help.

aksonov:
@smallsaucepan Yes, please.

[...]

PR #2532, opened 2017-10-22
Add PropTypes and update dependencies to work with react 16.0.0

daviscabral:
@aksonov the PR you asked for in #69.

### DIFF
--- a/Example/components/NavigationDrawer.js
+++ b/Example/components/NavigationDrawer.js
@@ -1,4 +1,5 @@
-import React, { PropTypes } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
 import Drawer from 'react-native-drawer';
 import { DefaultRenderer, Actions } from 'react-native-router-flux';
 
--- a/Example/components/TabView.js
+++ b/Example/components/TabView.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import {PropTypes} from "react";
+import PropTypes from 'prop-types';
 import {StyleSheet, Text, View, ViewPropTypes} from "react-native";
 import Button from 'react-native-button';
 import { Actions } from 'react-native-router-flux';
--- a/docs/REDUX_FLUX.md
+++ b/docs/REDUX_FLUX.md
@@ -104,7 +104,8 @@ Now you can access the current scene from any connected component.
 
 ```jsx
 // components/MyComponent.js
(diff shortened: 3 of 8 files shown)


daviscabral:
@aksonov could you please bump the version too?

aksonov:
Done

daviscabral:
Thanks.

ujwal-setlur:
This does not fully resolve things for me. I am on RN 0.49.3, React 16.0.0-beta.5. PropTypes has been deprecated from React and moved to prop-types. This PR addresses some of this, but not completely. react-native-router-flux depends on an old version of react-navigation-experimental which also PropTypes from react. It needs to be get fixed there as well.

daviscabral:
@ujwal-setlur #2541 address it - and also adds some performance improvements.

ujwal-setlur:
Thanks! Will check it out.

schlaegerz:
Checked out your branch, it still doesn't completely work. It appears there are a few remaining propTypes declared incorrectly

michaelstokes:
Hello,

Is this still a pending issue? 

I've noticed my app is crashing in release mode but works fine in debug?

Thanks very much!

ujwal-setlur:
I think it should be all fixed now

daviscabral:
@michaelstokes do you have more information about your issue? could you open a new issue with all the data from the template filled? tag me there please, so I can help you there.

michaelstokes:
Hi @daviscabral 

Thanks, I was able to address the issue by doing the following 

https://stackoverflow.com/questions/47603791/react-native-undefined-is-not-an-object-evaluating-react2-proptypes-string/47603838#47603838

Using react-codemon, run on your projects directory

jscodeshift -t react-codemod/transforms/React-PropTypes-to-prop-types.js myapp/

This will resolve updating proptypes for your source code and any node modules.

Re-installing node modules will require running the command again it seems.

Please follow instructions over at https://github.com/reactjs/react-codemod but essentially once you've installed jscodeshift, and you've downloaded the react-codemod files as above for the various commands such as transforming React.propTypes to PropTypes and so on.

myapp/ here corresponds to my entire application project directory so the entire react-native project.

Secondly, consider the project dependencies you're using. For example in my case I was using react-native-router-flux at version 0.38.0, the app would work in debug mode, but crash in release. I attempted to upgrade to the latest beta version of 4.0.0-beta.24 and the problem was resolved.

PR #2574, opened 2017-11-01
bind early. For faster render

kesha-antonov:


src/navigationStore.js
```diff
@@ -294,12 +298,15 @@ function createWrapper(Component, wrapBy, store: NavigationStore) {
           store.deleteRef(originalRouteName(navigation.state.routeName));
         }
       }
+      onRef(ref) {
```
ovr: `onRef(ref) {` -> `onRef = (ref) => {` you can use this instead of bind manually inside `constructor`, what do you think?

kesha-antonov: I'm not sure about performance. Which is faster?

But both will be faster than it's now since function for ```ref``` re-created on every render.

ovr: https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAECC0G8BQBIAZgVwHbAC4EsB7LAIXywBMSBPWAJzsIHdoBeaACgEo2A-RJNCHRBwgL5IJSUJBgkBQ4MQi46GPITrcF0FLgAW-CADpcAUxVtoBo6Yu5jAI3IUONiFwDcoiUNHmVbQQJKQA3MDprAAYrAAczOjRNAFswHDNjLGZubyQkyI4QM1xofCsoz1LoAB5oAEYoxsr8AGoWnmRhaCwzFlgcySQkcMjcOriE_NT0zOyvIfzOIpKy9gqq2oam0raO0SEelhIBsIjrACYJxJS04AyspgHpZUIi4xBCAHMOACI6gC4ftAWpwxtAALTRHggoHJfAgEBGMxKSgQAA00HOgOBoMukLGXC4QA&debug=false&circleciRepo=&evaluate=true&lineWrap=true&presets=es2015%2Ces2015-loose%2Ces2016%2Ces2017%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&targets=&version=6.26.0

image

1:0.4900000000000091 vs 2:0.3149999999999409

kesha-antonov: @ovr  bind is faster?

PR #2584, opened 2017-11-02
Upd deps: react-navigation to beta-15, rm RN-button, upd mobx & mobx-react

kesha-antonov:


aksonov:
Please update also yarn.lock, also react-navigation.beta.17 was published... Maybe we could use something like '*' to install latest beta automatically?

kesha-antonov:
@aksonov Ok! one moment.

There was style issue with navBar in beta.16. react-navigation team fixed it

Also I'm not confident enough in react-navigation usage or RNRF source code to be sure if it's a good idea or we should manually check every release of react-navigation before update.

kesha-antonov:
What do you think about to update mobx & mobx-react?
screen shot 2017-11-04 at 21 03 33

kesha-antonov:
I've looked through mobx changelog. Nothing breaking

PR #2605, opened 2017-11-08
fix(typings): add missing properties and in general fix typings

sondremare:
Significant change is that TabProps no longer extend SceneProps, as Tabs do not need to have a component set. Was unable to add which properties should be inherited from SceneProps to TabProps due to the fact that there is no code in Tabs.js


aksonov:
Could you resolve conflicts?

sondremare:
Done @aksonov

PR #2691, opened 2017-12-08
Update tabBarOnPress handler to use react-navigation beta.20 contract

rsailor:
React Navigation version requirement `"^1.0.0-beta.19"` in `package.json` as of today installs version `1.0.0-beta.21` on new installs of React Native Router Flux.

Version `1.0.0-beta.20` introduced a new contract for tab onPress callbacks.

                        onPress={() =>
     -                  onPress ? onPress(scene, jumpToIndex) : jumpToIndex(index)}
     +                  onPress
     +                    ? onPress({ previousScene, scene, jumpToIndex })
     +                    : jumpToIndex(index)} 

This breaks `backToInitial`. It will break any other part of the code that relies on the callback that React Navigation passes; I don't see any other place where this happens, but am welcome to updates to this PR to address additional callback changes.

This PR fixes the callback handling in `navigationStore` and restores the functionality of  `backToInitial`.

An alternative to this PR is to lock the React Navigation version to `1.0.0-beta.19`.


aksonov:
Looks great!

PR #2745, opened 2017-12-28
Drawer fixed, faulty closeDrawer removed, androidBack support added

vibhavagarwal5:
- Drawer had an error, fixed it.
- There was a closeDrawer command in DrawerContent.js which wasnt working so removed it
- Android back wasn't supported so added that.
- Before:
screenshot_20171228-212206
- Before:
screenshot_20171228-212712
- After:
screenshot_20171228-212514
- After:
screenshot_20171228-212728


vibhavagarwal5:
@aksonov do check the request.
this solves many problem for beginners (Faced in the example app)

PR #2769, opened 2018-01-08
Component prop should be optional

erkrapide:
The interface SceneProps should have the component prop set to optional (?), because it won't build in typescript when using a Nested Scene for Tabs. The docs claim to have this prop as and (semi)-optional.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -26,7 +26,7 @@ export var Scene: SceneStatic;
 export type Scene = SceneStatic;
 interface SceneProps extends React.Props<Scene> {
     key: string;
-    component: React.ComponentType<any>
+    component?: React.ComponentType<any>
     back?: boolean;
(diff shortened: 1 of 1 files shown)


aksonov:
I can approve it, but probably it is better to make component required for `Scene` and use other container types for containers - `Tabs`, `Stack`, `Drawer`, does it make sense?

erkrapide:
I see, you are absolutely right. I missed the `Tabs (<Tabs> or <Scene tabs>)` part. I was using `<Scene tabs>` construction.

dernise:
Right, but shouldn't we deprecate the old syntax and update the docs to remove all the <Scene key="root"> ? It's kind of confusing for typescript users

erkrapide:
@aksonov could you bump the version, please?

PR #2853, opened 2018-02-01
Compatibility with RN 0.52.1 for RNRF v3

DavidPineda:
in my company I was asked to update our app to the latest version of RN, to achieve it I found many errors with the V4 of this library also we did not want to release to production with beta libraries, so I modified some things from version 3.38. 0 of your library to work with RN 0.52.1, basically the changes are to add the import of the PropTypes from your own package.

It could be something like a version 3.38.1, I do this pull requests in case it would serve someone in the future.


aksonov:
Could you check conflicts?

DavidPineda:
Conflicts resolved

duhseekoh:
So, looks like there was never a version published to npm that includes this compatibility? Latest I see for 3.x is 3.43.0 which was released in October, months before this was merged.

Can we get this published as 3.44.0?

PR #2956, opened 2018-03-22
Update mobx & mobx-react to latest version

syq7970:
For this issue:

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,8 @@
   },
   "dependencies": {
     "lodash.isequal": "^4.5.0",
-    "mobx": "^3.3.1",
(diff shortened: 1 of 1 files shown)


heralight:
A far more better option is to change these dependencies to peer dependencies ... 
Mobx Issue and related docs:
https://github.com/mobxjs/mobx/issues/1082

Mobx docs:
1. If there should be only one mobx instance (libraries should cooperate & react to each other)
Use peer dependencies

aksonov:
Sorry for the delay. Closing this ticket because latest version (based on React Navigation 2.x) doesn't use mobx

PR #2979, opened 2018-04-02
migrate mobx to peer and dev dependencies

heralight:
To avoid mobx update dependency everytime and to not force projects to upgrade with RNRF...

A far more better option is to change these dependencies to peer dependencies ...
Mobx Issue and related docs:
mobxjs/mobx#1082

Mobx docs:

If there should be only one mobx instance (libraries should cooperate & react to each other)
Use peer dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,6 @@
   },
   "dependencies": {
     "lodash.isequal": "^4.5.0",
-    "mobx": "^3.3.1",
-    "mobx-react": "^4.3.4",
     "opencollective": "^1.0.3",
     "path-to-regexp": "^2.1.0",
     "prop-types": "^15.6.0",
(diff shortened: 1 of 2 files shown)


aksonov:
Does it require different setup process for end user who doesn't use mobx?

heralight:
Sorry, I missed your question...

In recent versions peer dependencies will not be installed automatically, but user will be pointed out that he should install them. So he has to install `yarn mobx mobx-react`. But it is the only correct thing to do, otherwise he cannot (properly) update mobx versions etc without you releasing a different version from rnrf.

aksonov:
What about following idea - clone this repo and use react-native-router-mobx with peer mobx dependencies (who are using mobx in their projects) and use normal dependencies for this project (for everybody who don't use mobx) ?

heralight:
I already done this (my fork, my push, as you can see in this push request, I use it directly in my projects). but how it's done actually in  react-native-router-mobx with  normal mobx dependencies, it's clearly not the right direction. and I don't understand the limitation for everybody who don't use mobx, it's one line to set up to have flexibility and no more mobx maintenance...

aksonov:
Ok, let's try your way.

PR #3000, opened 2018-04-17
Change behaviour of static navigationOptions property.

charles-salmon:
### Current behaviour
* When a `navigationOptions` static property is set on a component used by a `Scene`, this overrides all navigation options used by the `Scene`.

### New behaviour
* When a `navigationOptions` static property is set on a component used by a `Scene`, this overrides only those options contained in the `navigationOptions` property.

### Use case
* When setting a `title` using the `navigationOptions` property, this prevented a `back` prop added to a `Scene` from having any effect.
* This PR will be useful to help with transitions from using React Navigation.

### Example implementation of use case
**App.js (entrypoint):**
```
import { Router, Scene, Drawer, Stack } from 'react-native-router-flux';

import * as Screens from './Screens';

export default class Navigator extends Component {
  render() {
    return (
      <Router>
        <Scene key="root" hideNavBar>
          <Drawer key="main" contentComponent={Screens.MenuDrawer}>
            <Scene key="home" component={Screens.Home} initial />
            <Stack key="myProfileStack">
              <Scene key="myProfile" component={Screens.MyProfile} initial />
              <Scene
                key="changeProfilePicture"
                component={Screens.ChangeProfilePicture}
                back
              />
            </Stack>
          </Drawer>
        </Scene>
      </Router>
    );
  }
}
```

**Screens.js:**
```
import { Text, View, TouchableOpacity } from 'react-native';
import { Actions } from 'react-native-router-flux';

export class MenuDrawer extends Component {
  render() {
    return (
      <View>
        <TouchableOpacity onPress={Actions.home}>
          <Text>Home</Text>
        </TouchableOpacity>
        <TouchableOpacity onPress={Actions.myProfile}>
          <Text>My Profile</Text>
        </TouchableOpacity>
      </View>
    );
  }
}

export class Home extends Component {
  render() {
    return <Text>Currently at Home</Text>;
  }
}

export class MyProfile extends Component {
  render() {
    return (
      <View>
        <Text>Currently at MyProfile</Text>
        <TouchableOpacity onPress={Actions.changeProfilePicture}>
          <Text>Navigate to Change Profile Picture</Text>
        </TouchableOpacity>
      </View>
    );
  }
}

export class ChangeProfilePicture extends Component {
  render() {
    return <Text>Currently at ChangeProfilePicture</Text>;
  }
}
```

### Description of implementation
* Prior to change: If you open the draw menu, navigate to "My Profile", and then navigate to "Change Profile Picture", you will observe that this page has a menu icon, despite having been passed a `back` prop in the `Navigator`.
* After change: If you open the draw menu, navigate to "My Profile", and then navigate to "Change Profile Picture", you will observe that this page has a back icon, in-line with the expectation of providing a `back` prop in the `Navigator`.


aksonov:

Issue #2932, opened 2018-03-13
<Tabs> backToInitial and tabBarOnPress interaction


blah1234:
### Version
- react-native-router-flux v4.0.0-beta.28
- react-native v0.53.0

### Actual behaviour
I've found that if the "backToInitial" property is set for Tabs, then any function defined for the "tabBarOnPress" property is not called.  

### Expected behaviour
After browsing through the source code a bit, I see that "backToInitial" itself is implemented as a function set to "tabBarOnPress".  I would have expected that any function users pass into the "tabBarOnPress" property be chained together with the "backToInitial" function if both "backToInitial" and "tabBarOnPress" are set for Tabs.

At the very least, the API documentation should notify users of the current behavior.

aaron25mt:
Created a fix for this, will make a PR.
See [my repo here](https://github.com/aaron25mt/react-native-router-flux/tree/dev) if you need the fix ASAP.

aksonov:
Please try to reproduce it with Example project and latest version 4.0.0-beta.40. Feel free to open if the issue still exists

daviscabral:
PR for better documentation to avoid this confusion for other users is welcome. But yeah, looking at the code I saw that if you define `backToInitial` - it's not possible to define `tabBarOnPress` - and keeping both handlers might be even worst (and harder to debug). I believe a warning would be fine - telling the user that the `tabBarOnPress` defined will be overwritten due the `backToInitial` flag defined.

daviscabral:
I am pushing a PR with this change - I believe it does not hurt. :-)

PR #3012, opened 2018-04-25
Allow developer to specify a tabBarOnPress function and set backToInitial to be true for tabs

aaron25mt:
See issue #2932

Previously, if backToInitial was set, any function set in tabBarOnPress would not be called. In this commit, I checked if the developer specified a function and set backToInitial to be true; if so, we will run the code for backToInitial/set tab child to initial Scene and then call the user specified tabBarOnPress function.


aksonov:
Looks great, could you also update docs and regenerate navigationStore again?

aksonov:

Issue #3008, opened 2018-04-23
evaluating addListner


Pitty91:
### Version
4.0.0-beta.30 

- react-native-router-flux 4.0.0-beta.27
- react-native 0.55.3

 I am facing  the issue #2808 
so I upgraded my router-flux version from 0.27 to 0.30

now facing the issue #2865 is there any proper documentation for this issues

![screenshot_1524466591](https://user-images.githubusercontent.com/17979573/39111124-918c1ffa-46f1-11e8-8dfa-f0d1336b49e1.png)

Pitty91:
I solved my issue by removing ^ from the _package.json_ 

   ` "react-native-router-flux": "4.0.0-beta.27",`

so I am closing it

Issue #3006, opened 2018-04-22
Example app doesnt run


ranadeep47:
### Version
Tell us which versions you are using: 

- "react-native-router-flux": "4.0.0-beta.30"

### Expected behaviour
The example app should run fine

### Actual behaviour
I ran `yarn install` and `react-native run-android` inside the `Examples` folder and the app fails with an error
![image](https://i.imgur.com/2nwNRMT.png)

serjek:
please see https://github.com/aksonov/react-native-router-flux/issues/2976

aksonov:
4.0.0-beta.31 fixes it, please check latest Example

Issue #2976, opened 2018-03-30
Examples Broken


duro:
### Version
Tell us which versions you are using: 

- react-native-router-flux v4.0.0-beta.30

Environment:
  OS: macOS High Sierra 10.13.3
  Node: 8.2.1
  Yarn: 1.3.2
  npm: 5.3.0
  Watchman: 4.7.0
  Xcode: Xcode 9.1 Build version 9B55
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.3.0-alpha.1 => 16.3.0-alpha.1
  react-native: 0.54.2 => 0.54.2

### Expected behaviour

Following example steps should successfully build example locally

### Actual behaviour

I get this error in simulator

![iphone_6_-_ios_11_1](https://user-images.githubusercontent.com/17632/38157379-38c3f6ea-343b-11e8-866e-ce00f5179c36.jpg)

### Steps to reproduce
For non-obvious bugs, please fork this component, modify Example project to reproduce your issue and include link here.

```
# Get the code
git clone https://github.com/aksonov/react-native-router-flux.git
cd react-native-router-flux/Example

# Install dependencies
yarn

# Run it
react-native run-ios
```

asciifaceman:
There is unfortunately some crazy things going on upstream, https://github.com/react-navigation/react-navigation/issues/3416


To get you going, and just for my own curiosity, can you try to uninstall the react-native-router-flux package in the project and then install the version pinned `4.0.0-beta.28` or `4.0.0-beta.22`

IronTony:
Tried with `4.0.0-beta.28` and it works

aksonov:
4.0.0-beta.31 fixes it as well

Issue #2897, opened 2018-02-21
Undefined is not a function (evaluating 'addListener') 


Dorentin1997:
### Version
Tell us which versions you are using: 

- react-native-router-flux ^4.0.0-beta.24
- react-native 0.50.3
- react-navigation ^1.0.0-beta.23

### Expected behaviour
Application should run normally

### Actual behaviour
It crashes after i just run "npm install"

TypeError: undefined is not a function (evaluating 'addListener')

This error is located at:
    in CardStack (at CardStackTransitioner.js:67)
    in RCTView (at View.js:113)
    in View (at Transitioner.js:142)
    in Transitioner (at CardStackTransitioner.js:19)
    in CardStackTransitioner (at StackNavigator.js:39)
    in Unknown (at createNavigator.js:13)
    in Navigator (at createNavigationContainer.js:227)
    in NavigationContainer (at Router.js:67)
    in App (at Router.js:88)
    in Router (at Routes.js:47)
    in Routes (at App.js:126)
    in App (at renderApplication.js:35)
    in RCTView (at View.js:113)
    in View (at AppContainer.js:102)
    in RCTView (at View.js:113)
    in View (at AppContainer.js:122)
    in AppContainer (at renderApplication.js:34)
<unknown>
    D:\Projects\Wertheim2017\SmartSafeMobileApplicaiton\node_modules\react-navigation\src\getChildEventSubscriber.js:59:16
getChildEventSubscriber
    D:\Projects\Wertheim2017\SmartSafeMobileApplicaiton\node_modules\react-navigation\src\getChildEventSubscriber.js:58:49
_getScreenDetails
    D:\Projects\Wertheim2017\SmartSafeMobileApplicaiton\node_modules\react-navigation\src\views\CardStack\CardStack.js:106:10
render
    D:\Projects\Wertheim2017\SmartSafeMobileApplicaiton\node_modules\react-navigation\src\views\CardStack\CardStack.js:219:47
proxiedMethod
    D:\Projects\Wertheim2017\SmartSafeMobileApplicaiton\node_modules\react-proxy\modules\createPrototypeProxy.js:44:35
finishClassComponent
    D:\Projects\Wertheim2017\SmartSafeMobileApplicaiton\node_modules\react-native\Libraries\Renderer\ReactNativeFiber-dev.js:1667:86
updateClassComponent
[...]

PR #3032, opened 2018-05-13
Add listeners to navigation (as part of react-navigation 1.0.0 breaking changes)

daviscabral:
After trying to work in a fix for a very old issue related to a misaligned title in the `NavBar` - I really got bothered that I couldn't run the `Example` project - and also even use the most recent project.

After a few hours digging in the code - noticed the missing listener in the `navigation` props. It still might have some bad side effects - since I am only forwarding the existent method to it.

This PR might be the fix for #2897, #3006, #2976, #2865 and #3008.

EDIT: Also a few major/minor version updates of some packages (npm is still whining about lots of security issues - `npm audit` to get a better report about it).


aksonov:
Looks great!

Issue #3026, opened 2018-05-08
popTo causes an infinite loop when route is not found


Jarred-Sumner:
### Version
Tell us which versions you are using: 

- react-native-router-flux v4.0.0 beta 27 (v3 is not supported)
- react-native v0.5.5

### Expected behaviour
- popTo returns `null` or throws an Error of some kind

### Actual behaviour
When calling `Actions.popTo` on a route that is not in the navigation stack, it infinite loops (inside of the `Reducer`)

Additionally, I expected it to "just work" given a parent route. More context:
- I have some tabs and some modals of $N depth. One of the modals are the current scene. I want to dismiss the modal. Going back just once might not be enough. The route I want to go back to is whichever was last selected on a tab. To do that, I wrongly assumed I could pass `Actions.popTo` the tabs route and it would "just work", instead it infinite loops. 
- My solution is to manually track what the last child route of the tab bar that was active, and `popTo` that child route

### Steps to reproduce
For non-obvious bugs, please fork this component, modify Example project to reproduce your issue and include link here.
1. Open your app
2. Call `Actions.popTo("AnyRouteThatDoesntExist")
3. Infinite loop

daviscabral:
@Jarred-Sumner I am able to reproduce the error - but it seems to be related to `react-navigation`. But I believe something can be done over it to avoid that. I will spare some time this week to investigate this issue. Thank you.

daviscabral:
Alright - I believe I found the cause of the issue. When reaching the top level, the state is always the same - never finding the undefined route. That might also happen with a defined route in a different stack level. 

Adjusted to verify if the current state is the same as the top level from the current router - if does, just break the loop.

AlexDM0:
I solved this by overloading the popTo in the reducer that you give to the Router.createReducer.

```js
export const reducerCreate = (params) => {
  const defaultReducer = Reducer(params, {});
  return (state, action)=>{
    if (action && action.type == "REACT_NATIVE_ROUTER_FLUX_POP_TO") {
      // check if we can see the key in the list of items, if not, do a back.
      let popCount = 0;
      let success = false;
      // search throught all routes in the state, starting from the top, for one that matches the route name
      for (let i = state.index; i >= 0; i--) {
        if (state.routes[i].routeName == action.routeName) {
          // we found it!
          success = true
          break;
        }
        // how many pops do we need for this operation?
        popCount++;
      }
      if (success) {
[...]

PR #3043, opened 2018-05-20
popTo causes an infinite loop when route is not found (as seen on #3026)

daviscabral:
I am not sure about the solution here. Right now it is just assuming that it does not exist and returning to the top level route existent. 

The problem with this approach is that might allow calls to inexistent routes pass unknown. 

What do you think here guys? 

1) Add a warning about the unknown route;
2) Thrown an error if reaches the top level without finding a `routeName` that matches;
3) Keep like it is in this PR, just redirects to the top level.

(This PR is about the problem reported in #3026)


daviscabral:
@aksonov any objections about getting this merged?

aksonov:
Strange, I didn't received any email about that, sorry. will review now

PR #3114, opened 2018-07-05
Fix prevScene to update on scene change only

bilalsyed001:
When using Action.refresh prevScene is updated to same as current scene which may not be desirable.


aksonov:

PR #3131, opened 2018-07-17
fix bug where backToInitial Flag did not work for cloned scenes

nhayfield:
Also seems to fix this issue from stackoverflow:
https://stackoverflow.com/questions/47352572/react-native-router-flux-reset-stacknavigation/51388834#51388834


aksonov:

Issue #3068, opened 2018-06-06
Adopt RNRF for latest React Navigation (v2+)


aksonov:
Current 4.0.0-beta.31 depends from react-navigation 1.5.6. It would be great to migrate to latest v2. `createNavigator` API is changed and some other breaking changes were made..

daviscabral:
I have this partially done - when finished I can push a PR and reference it here.

gyf19:
I want to know what version of react-native is required after the upgrade?

daviscabral:
@gyf19 I'll confirm it as soon as I get back to contribute here - probably end of next week. Thank you.

AlexDM0:
I'm really looking forward to an integration with the new versions since the following bug is very visible in our app: https://github.com/react-navigation/react-navigation/issues/4185

It's fixed in 2.5.5 :)

aksonov:
Unfortunately it seems to be quite hard... I started and stuck here: https://github.com/react-navigation/react-navigation/issues/4481

As you can see react-navigation starts recommending to store state within their library only and will not support (?) external management of state with Redux, MobX, etc. (as well as RNRF), etc. starting next major version. I will think what I can do.

aksonov:
I found one easy and one complex way. First is to remove state management from RNRF, second is to build own navigators and use them instead of react-navigation ones, use only views from react-navigation. I'm working on first way while probably second way may be implemented at the future (or maybe not).

Feel free to check my PR (Some links work, some not - login/lightbox/modal don't work yet)

aksonov:
@daviscabral what do you think about my comments above and my PR? There are still a lot of things to to do, maybe you can help here (new tab navigator, our custom actions like popTo, etc)....

daviscabral:
@aksonov Great - I've been thinking about removing state management from RNRF - but that would make the project name a bit out of sync with it. I'll check your PR and start to hack on missing stuff - let me know things that you would prefer me to look into first.

aksonov:
@daviscabral Could you clarify - "make the project name a bit out of sync with it." ? RNRF name?
You may start from converting navigators and restore original Example demo, fix eslint errors or restore original actions like POP_TO, checking all other actions like RESET, REPLACE - they may not work..., etc.

daviscabral:
@aksonov sure - yeah - when I was trying to make it to work with `react-navigation@2` here - ended being more like a utils than something like it is now. But checking your PR, that's a way better than I thought. 

[...]

PR #3141, opened 2018-07-22
migration to react-navigation v2.x (#3068)

aksonov:
- [x] Converting navigators
- [x] Restore original Example demo
- [x] Fix eslint errors (#3142 @daviscabral)
- [x] Restore original actions like POP_TO checking all other actions like ~RESET~ (fixed in 49b8991), ~REPLACE~ (fixed in e8085cd)
- [x] Remove mobx
- [x] Fix onEnter/onExit handlers and custom onNavigationStateChange handler (@aksonov)
- [x] Fix navTransparent (@aksonov)
- [x] verify deep linking


daviscabral:
@aksonov I've brought your list here - and placed my username beside the item I am planning to tackle first.

Example/package.json
```diff
@@ -12,16 +12,16 @@
     "mobx": "^3.2.1",
     "mobx-react": "^4.2.1",
     "prop-types": "^15.5.10",
-    "react": "16.4.1",
-    "react-native": "0.56.0",
```
daviscabral: Any reason to revert this here? Something that I might help? I was thinking about using the last version for everything. I've spent a couple of days in the past week just to make a few projects to work with these versions.

aksonov: Yes, please try to fix. I've got some babel-related error, connected with stage 0 or something like that.

daviscabral: Cool - I think I know how to fix them. This might be related to the use of babel 7 (I had to upgrade a few other deps in order to make it to work - checking it).

daviscabral: Got example working with latest `react` and `react-native`. 

ezgif-5-da09ec92f0

daviscabral: Also from:
```
✖ 538 problems (535 errors, 3 warnings)
```

To:
```
✖ 136 problems (133 errors, 3 warnings)
```

daviscabral:
I believe you are using the plugin for prettier only - there is a better plugin with support to eslint + prettier on VSCode - better to use that one instead.

aksonov:
Could you say name of plugin? There is a lot..

daviscabral:
@aksonov this one: 

https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

You have to enable the ESlint integration: 

```
prettier.eslintIntegration: true
```

Also, instead of `prettier`, use `prettier-eslint`, so it will apply it in the right order.

daviscabral:
👏 👏 👏 👏

aksonov:
@daviscabral I'm done for today. If you have time please fix custom actions, as well replace action, also I found missed labels for two tabs, etc. Btw, I introduced easier custom navigator implementation now - 'renderers', i.e. views for custom navigators, please check OverlayRenderer and LightboxRenderer.

daviscabral:
@aksonov sure - the custom actions are almost done. The missing labels are the next thing to do.

aksonov:
Oh, also custom onNavigationStateChange handler.

daviscabral:
Got busy with work and couldn't finish what I've started - but planning tackle that later 👍

src/navigationStore.js
```diff
@@ -471,25 +466,69 @@ function uniteParams(routeName, params) {
   return res;
 }
 
+const defaultSuccess = () => {};
+const defaultFailure = () => {};
+
```
daviscabral: My bad here for removing it - they were not being used before.

aksonov: No worries, you just fixed eslint errors. It was used by removed setState, now I re-enabled handlers.

aksonov:
@daviscabral Please check deep linking if you can, it may not work now..

aksonov:
@daviscabral any ETA for REPLACE action?

daviscabral:
Checking it now - back with ETA in a few.

daviscabral:
I believe there are still more things besides what is listed in the PR todo list - do you recall anything else @aksonov?

daviscabral:
Sure - good with that. I'll check about the missing action then.

aksonov:
Oh, missed texts for two tabs within Example, do you know what it could be?

daviscabral:
Checking.

daviscabral:
@aksonov can you elaborate on this to me?

```js
            // go to first screen of the StackNavigator with reset
            // navigation.dispatch(NavigationActions.reset({
            //   index: 0,
            //   actions: [NavigationActions.navigate({ routeName: tab.route.routes[0].routeName })],
            // }));
            // go to first screen of the StackNavigator without reset
            for (let i = 1; i < scene.route.routes.length; i += 1) {
              navigation.dispatch(NavigationActions.back());
            }
```

Why are we using multiple back calls instead of just one single reset there?

aksonov:
@daviscabral Because reset was buggy and didn't work for sub-tree, feel free to remove it

PR #3147, opened 2018-07-26
Navigation and replace restored

daviscabral:
@aksonov ended reverting what I was doing and simplifying by using the original `react-navigation` actions.


daviscabral:
Still checking about it - taking a look in the `POP_TO_TOP` to see if we could do something similar.

PR #3175, opened 2018-08-13
Upgrade react-navigation@2.11.2

Agby:
Upgrade react-navigation@2.11.2 to prevent break jest test case


daviscabral:
Sorry - I thought I had commented here. Please, update the yarn.lock file so we can have it merged :)

PR #3179, opened 2018-08-14
Upgrade react-navigation@2.11.2

Agby:
Re-open Upgrade react-navigation@2.11.2 to prevent break jest test case
Also update `yarn.lock`


Agby:
Thank you :)

PR #3191, opened 2018-08-18
Releasing v2.3.14

daviscabral:
Branch cleanup and README updates.


daviscabral:
After digging a bit more, makes no sense to do this PR and mess around with old code like that. Closing this in favor of handling with issues and PRs that are pilling every day.

PR #3212, opened 2018-08-24
add Statusbar marginTop for navTransparent

rikochet:
When using navTransparent feature, the headerStyle is removed causing the navbar to be obstructed by the Statusbar. 

This fix adds marginTop to the headerStyle to that the navbar is aligned below the Statusbar.

### DIFF
--- a/src/navigationStore.js
+++ b/src/navigationStore.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Image, Animated, Easing } from 'react-native';
+import { Statusbar, Image, Animated, Easing } from 'react-native';
 import {
   createBottomTabNavigator,
   createMaterialTopTabNavigator,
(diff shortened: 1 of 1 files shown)


daviscabral:
Do you mind showing how is that behaving on Android and iOS before and after? I am trying here and not being able to see the problem or to confirm that this change is not going to break anything else for other people.

rikochet:
This issue only seems to affect Android.. You can see the difference before and after my patch in the attached illustration. iOS seems to display fine both with and without a StatusBar height..

patch

kylanhurt:
Are we sure that this was an appropriate fix? My transparent header NavBar was positioned fine before this update and now it appears as if it is pushed down by 24px on Android. According to this other GitHub issue it appears I'm not the only one having this problem.
https://github.com/aksonov/react-native-router-flux/issues/3291

Unfortunately since the headerStyle is overwritten for a `navTransparent` of `true` it means that I cannot manually change the `marginTop` back to 0.

Bang9:
@daviscabral This is not good..
I'm not using status bar as absolute position in my app.. this merge occur error to me
Moreover, since the header itself set margin top, the user cannot set the styling through navigationBarStyle prop

PR #3214, opened 2018-08-25
Add support for deprecated Tabs component

daviscabral:
Add support to deprecated Tabs component to allow `swipeEnabled` and `animationEnabled`. Also updates the API docs to inform that those options are not available to `Tabs` anymore.


daviscabral:
I believe it's related with the versions - I spent some time tracking the dependency that could be causing it - but without success yet.

daviscabral:
That's weird - locally it's passing. Checking what is going on.

daviscabral:
@aksonov tests fixed - just lint updates.

PR #3228, opened 2018-08-29
Add title to StackProps type definition

vanb:
The Stack component supports a title property. This updates typescript definitions to match.


daviscabral:
Thank you @vanb!

Issue #3236, opened 2018-09-02
Can't pass props to scene from Action.key(props). Think this way should work too.


vygandas:
Hello,
I can't send props to another scene. Was looking into the docs but can't find that could help.

### Version
Tell us which versions you are using: 

- react-native-router-flux 4.0.1
- react-native 0.55.4

### Expected behaviour
Action.scene_key(props) shout pass props to the component

### Actual behaviour
No props in the component. I see only Router specific props printed out:
![screen shot 2018-09-02 at 10 52 12](https://user-images.githubusercontent.com/723200/44953485-9b65b580-ae9e-11e8-8253-423ea3a23a11.png)

### Steps to reproduce
For non-obvious bugs, please fork this component, modify Example project to reproduce your issue and include link here.

App.tsx
```
<...>
<Router>
    <Stack hideNavBar>
        <Tabs key="root" showLabel={false} swipeEnabled={true} lazy={true}>
            <Scene key="feed" component={FeedScreen} iconName="ios-people" icon={TabIcon} navBar={FeedNavBar} />
            <Scene {...this.props} key="selfie" component={SelfieScreen} iconName="ios-camera" icon={TabIcon} iconSize={30} />
            <Scene {...this.props} key="chats" component={ChatsScreen} iconName="ios-chatboxes" icon={TabIcon} title="Conversations" />
        </Tabs>
        <Scene key="settings" component={SettingsScreen} iconName="ios-settings" icon={TabIcon} modal={true} hideNavBar={false} title="Settings" />
        <Scene {...this.props} key="messages" component={MessagesScreen} iconName="ios-settings" icon={TabIcon} modal={true} hideNavBar={false} />
    </Stack>
</Router>
<...>
```

FeedItemBox on click action
```
<...>
Actions.selfie({
    target: 'chat',
    taget_user_id: this.props.item.uid,
    target_user_name: this.props.item.name,
});
<...>
```

Selfie component that should get props
```
<...>
render() {
    console.error(this.props);
    return (
        <View style={styles.container}>
            <Text>
                {this.props.target === 'feed' && 'Post your selfie!'}
                {this.props.target === 'chat' && `Say hello to ${this.props.target_user_name} with your selfie!`}
            </Text>
                
            <....>

        </View>
    );
}
<...>
```

daviscabral:
Where is how is `Actions.selfie` being called?

I noticed that with the last release, a lot of issues might be happening due the wrong version match between a lot of dependencies.

[...]

Issue #3218, opened 2018-08-27
Actions.refresh renderRightButton has no effect


jamesholcomb:
### Version
Tell us which versions you are using: 

- react-native-router-flux v4.0.1 (v3 is not supported)
- react-native v0.56.0

### Expected behaviour
The navbar refreshes with the new right button component

### Actual behaviour
No effect

### Steps to reproduce
For non-obvious bugs, please fork this component, modify Example project to reproduce your issue and include link here.

````
  static onEnter = () => {
    const { wrappedInstance } = Actions.refs.profile

    // this works
    Actions.refresh({
      rightTitle: "Logout",
      onRight: wrappedInstance.onLogout
    })

   // this has no effect
   Actions.refresh({
      renderRightButton: () => <MoreButton onPress={() => alert('hi')} />
   })
  }
````

jamesholcomb:
Here is how I worked around it

````
        <Scene
          back
          key={routes.PROFILE}
          component={ProfileScene}
          renderRightButton={
            <MoreButton
              onPress={() => Actions.refs.profile.wrappedInstance.onMorePress()}
            />
          }
          title="Profile"
        />
````

pierol:
I made an example code to show you what i mean go here: https://github.com/pierol/rnrf-example it's an EXPO example try to open it in a ios simulator. Make login inside the app with these credentials: pierol and pw: giulia09. You have to swtch the version of react-native-router-flux from 4.0.0.beta.27 and 4.0.1 to see the two different behaviours. There is a console log in the DayList componenetWillMount method this component is part of the first Scene. After the Login is successfully i will make an Actions.choorHour() call to return to the first Scene, the one with the Daylist component. Only using 4.0.0.beta.27 will reload the component DayList when i return to the main scene from the Login form.

daviscabral:
Thank you for the update about the workaround. This seems to be sort of a duplicate of #3188. I'll keep it marked as bug and work on both in the same PR.

aksonov:
@jamesholcomb It looks like a hack and probably just not supported by React Navigation 2.x (its SET_PARAMS action). Instead you should define one renderRightButton that changes its content depending from navigation state. `Actions.refresh` should be used just to update navigational state (i.e. pass values not functions)

jamesholcomb:
> Actions.refresh should be used just to update navigational state (i.e. pass values not functions)

[...]

Issue #3206, opened 2018-08-23
after update to Release 4.0.2 props not working...


lordenas:
### Version
Tell us which versions you are using: 

- react-native-router-flux v 4.0.2 (v3 is not supported)
- react-native ~0.55.2

### Expected behaviour

login .js
 Actions.Account_list({ data: 'test!' })

accontlist.js
console.log(this.props.data) = undefined

### Actual behaviour
console.log(this.props.data) = test!

lordenas:
https://github.com/lordenas/reactNativeJfSerive

In the loginForm.js function myoutput() after a response is received, makes the transition to `Action.Account_list({token: this.state.token, userName:this.state.userName})`

But Account_list.js stopped seeing these props

lordenas:
Something went wrong... I went back to yesterday's version and everything worked. Somehow ceased to be transferred to the props...

dextermb:
I am also noticing that my props are not being populated
```
<TouchableOpacity
	style={_Challenges.container}
	onPress={() => Actions['ChallengeDetails'](
		{
			...props,
			challenge: challenges[i]
		}
	)}
>
    // ...
</TouchableOpacity>
```

Output from `this.props` within `componentDidMount`:
```
Object {
  "animationEnabled": false,
  "drawerImage": 2,
  "drawerPosition": "right",
  "headerMode": "none",
  "init": true,
  "name": "_ChallengeDetails",
  "navigation": Object {
    "actions": Object {
      "closeDrawer": [Function closeDrawer],
      "dismiss": [Function dismiss],
      "goBack": [Function goBack],
      "navigate": [Function navigate],
      "openDrawer": [Function openDrawer],
      "pop": [Function pop],
      "popToTop": [Function popToTop],
      "push": [Function push],
      "replace": [Function replace],
      "reset": [Function reset],
      "setParams": [Function setParams],
      "toggleDrawer": [Function toggleDrawer],
    },
    "addListener": [Function addListener],
    "closeDrawer": [Function anonymous],
    "dangerouslyGetParent": [Function anonymous],
    "dismiss": [Function anonymous],
    "dispatch": [Function anonymous],
    "getChildNavigation": [Function getChildNavigation],
    "getParam": [Function anonymous],
    "getScreenProps": [Function anonymous],
    "goBack": [Function anonymous],
    "isFocused": [Function isFocused],
    "navigate": [Function anonymous],
    "openDrawer": [Function anonymous],
[...]

Issue #3197, opened 2018-08-21
hideTabBar is broken when used in nested stacks/scenes


R1ckye:
### Version
Tell us which versions you are using: 

- react-native-router-flux v4.0.2 (latest master)
- react-native v0.56.0 (Example App)

### Expected behaviour
hideTabBar on any Scene hides TabBar when that Scene is presented (as previously)

### Actual behaviour
TabBar keeps on showing

### Steps to reproduce
Slightly modified the Example app, added `tabs` to the root Stack, and added `hideTabBar` to the register Scenes, I have the same structure in my app. Previously using 4.0.0-beta.28, TabBar was hidden when the Scene with `hideTabBar` got presented, now it isn't.

          <Stack hideNavBar tabs key="root" titleStyle={{ alignSelf: 'center' }}>
            <Scene key="echo" back clone component={EchoView} getTitle={({ navigation }) => navigation.state.key} />
            <Scene key="launch" component={Launch} title="Launch" initial />
            <Stack key="customNavBar" hideTabBar titleStyle={{ alignSelf: 'center' }}>
              <Scene key="customNavBar1" title="CustomNavBar 1" navBar={CustomNavBar} component={CustomNavBarView} back />
              <Scene key="customNavBar2" title="CustomNavBar 2" navBar={CustomNavBar} component={CustomNavBarView} back />
              <Scene key="customNavBar3" title="Another CustomNavBar" navBar={CustomNavBar2} component={CustomNavBarView} back />
              <Scene key="hiddenNavBar" title="hiddenNavBar" component={CustomNavBarView} hideNavBar={true} back />
            </Stack>
            <Stack back backTitle="Back" key="register" duration={0} navTransparent>
              <Scene key="_register" hideTabBar component={Register} title="Register" />
              <Scene key="register2" hideTabBar component={Register} title="Register2" />
              <Scene key="home" component={Home} title="Replace" type={ActionConst.REPLACE} />
            </Stack>

Agby:
I had the same issue too

### Version
react-native-router-flux v4.0.1 

```
    <Drawer
      key="root"
      contentComponent={Sidebar}
      headerMode="none"
    >
      <Tabs key="myTabBar">
        <Scene key="tab_1" tabBarLabel="home">
          <Stack>
            <Scene key="home" component={Home} />
            <Scene key="blankPage" component={BlankPage} />
            <Scene key="listPage" component={ListPage} hideTabBar={true} /> // not work
          </Stack>
[...]

PR #3240, opened 2018-09-03
Refresh not working as expected in several cases

daviscabral:
Some small tweaks in the refresh method and a few new callbacks to allow some work around navigation events in order to get it updated. 

Components are mounted at once - from what I could see - still checking it.

This PR will solve #3236, #3218, #3206, #3197, #3188.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,26 @@
 
 [Full Changelog](https://github.com/aksonov/react-native-router-flux/compare/4.0.1...HEAD)
 
+**Implemented enhancements:**
+
+- New release of `react-navigation` is available \(v2.12.1\) [\#3222](https://github.com/aksonov/react-native-router-flux/issues/3222)
+- Upgrade react-navigation to v.2.12.1 [\#3223](https://github.com/aksonov/react-native-router-flux/pull/3223) ([daviscabral](https://github.com/daviscabral))
+
+**Fixed bugs:**
+
+- Can't pass props to scene from Action.key\(props\). Think this way should work too. [\#3236](https://github.com/aksonov/react-native-router-flux/issues/3236)
+- Actions.reset: There is no route defined for Key X [\#3194](https://github.com/aksonov/react-native-router-flux/issues/3194)
+- Stack of same scene not working anymore after update from 4.0.0-beta-28 to 4.0.1 [\#3176](https://github.com/aksonov/react-native-router-flux/issues/3176)
+- \[Fix \#3194 \#3176\] Adjust reset to work with nested routes [\#3237](https://github.com/aksonov/react-native-router-flux/pull/3237) ([daviscabral](https://github.com/daviscabral))
+- add Statusbar marginTop for navTransparent [\#3212](https://github.com/aksonov/react-native-router-flux/pull/3212) ([rikochet](https://github.com/rikochet))
+
 **Closed issues:**
 
+- How can i get the props after Actions.popTo in the previous scene? [\#3241](https://github.com/aksonov/react-native-router-flux/issues/3241)
+- There is no route defined for key when using Actions.reset\(\); [\#3234](https://github.com/aksonov/react-native-router-flux/issues/3234)
+- Proposal: RNRF 3.5 [\#3213](https://github.com/aksonov/react-native-router-flux/issues/3213)
 - How to pass the props in Tabs? [\#3203](https://github.com/aksonov/react-native-router-flux/issues/3203)
+- Navigation breaks after 4.0.0-beta.23 [\#3202](https://github.com/aksonov/react-native-router-flux/issues/3202)
 -  How can I hear the event come back? [\#3201](https://github.com/aksonov/react-native-router-flux/issues/3201)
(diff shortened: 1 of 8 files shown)


aksonov:
@daviscabral Is it ready to merge?

aksonov:
I think no, could you tell me what difficulties are?

daviscabral:
Just lack of time - helping a friend with some stress test in his server right now - after finishing it I'll push the changes. 

But what I am doing is to move the logic of `onEnter` from `componentDidMount` to the event listener `willFocus`. And the `onExit` to the `didBlur`. But I want to be sure that is really going to fix the issue - and also not break anything. 

I am also extending the test suite - so it's possible to require that every new PR that fixes a bug comes in companion of a test that helps it to come back (and also in future upgrades).

dextermb:
😲cant wait to see everything functional again, @daviscabral @aksonov

aksonov:
@daviscabral But how it will solve problems with passing params?

daviscabral:
@aksonov yes - that's going to fix that's as well.

daviscabral:
@aksonov after this - could we get a new version? I tried the last time but due the lack of permissions I was not able to publish - so I reverted a few things. Ended doing a little mess here while you was not around.

aksonov:
@daviscabral Sure, I'll publish new version once this PR is finished. Any ETA?

daviscabral:
Should be there in a few hours - just finishing a few tests here.

dextermb:
@daviscabral you've managed to resolve the param issue?

wheatleyjj:
Can't wait to get this update! Thanks @daviscabral

daviscabral:
Please, do not merge yet - I noticed a bug in the `isStatelessComponent` call - making it to pass a ref to the wrong kind of component. 

I've updated the examples to some of the cases that people complained - but to hide or pass props - it's like @aksonov said before - `wrap=false` might be needed - or you will have to provide the right scene key to `refresh` or `pop`.

daviscabral:
I would appreciate some help testing these changes by the people that were having the problems before we proceed. We might have to update docs too. Any voluntary?

dextermb:
@daviscabral I will test it now and see if it works in my testcase

dextermb:
@daviscabral Attempted to see if it worked with the example I posted a while ago (challenge). But it didn't seem to get passed down 🤕

dextermb:
@daviscabral I've created a [gist](https://gist.github.com/dextermb/522b0d0da94aa69b8996dd30c2fbaeda) with the example I gave, along with the routes file.

daviscabral:
Can you try this? (just to test if I am correct about the key reference that we need to keep now. If thta's the case - like it happened with React Navigation - it will require some work from everyone to adapt to it.

```
onPress={() => Actions.pop(
        {
          ...props,
          refresh: { challenge: challenges[i] },
          key: 'root'
        }
      )}
```

I believe you are defining the params but they are not passed from top to bottom like before. Also, check this:

```
onPress={() => {
   Actions.ChallengesDetails();
   Actions.refresh(
        {
          ...props,
          refresh: { challenge: challenges[i] },
          key: 'root'
        }
      )}
```

dextermb:
I've been trying to debug it myself from reading through code, but wouldn't the issues be happening with:
```
      if (!this[key]) {
        this[key] = new Function(
          'actions',
          'props',
          'type',
          `return function ${
            key.replace(/\W/g, '_') // eslint-disable-line no-new-func
          }(params){ actions.execute(type, '${key}', props, params)}`,
        )(this, { error: '', ...commonProps, ...props }, type);
      }
```
---

[Source](https://github.com/aksonov/react-native-router-flux/blob/dc/issues/3236-3218-3206-3197-3188-refresh-not-working-as-expected/src/navigationStore.js#L736-L745)

daviscabral:
I use this to help with the libs I help to maintain:

```bash
function rn-clean() {
  rm -rf android/build
  rm -rf ios/build
  watchman watch-del-all
  rm -rf $TMPDIR/react-*
  rm -rf $TMPDIR/haste-*
  rm -rf $TMPDIR/metro-*
  yarn start --reset-cache
}
```

```
$ cp [lib-path:react-native-router-flux(in this case)]/src/*.js [app-path]/node_modules/react-native-router-flux/src/
$ yarn install # in the project root 
$ rn-clean
```

For your case - probably worth also remove `node_modules`.

daviscabral:
This is the piece that magically generates the methods with scene names. But it calls execute - with type `jump` for tabs - or `push` for everything else.

dextermb:
@daviscabral Alright, ive cloned down my app again. `npm i` then going to build and test with Expo

dextermb:
@daviscabral Isn't that magic function the point where it degregates props into the other functions then down to react-navigation?

daviscabral:
You can place a `console.log` inside the `execute` - and you will see that the params get there. The problem seems to be that is not possible to put a lot of nested scenes like before without wrapping it with a navigator to manage this. This is what I am trying to confirm.

dextermb:
@daviscabral 😛 im trying to confirm it for you but my app doesn't want to play nice after changing versions

dextermb:
Was unable to get the app working again after trying different things, so i'm transplanting things into an application that uses React Navigation for now

daviscabral:
I am getting this in - I can't reproduce that - but it fixed a few other cases and add more examples.

PR #3244, opened 2018-09-05
Updating with live

RodrigoCS:


daviscabral:
Sorry - this does not seems to be valid for master. I just created a branch for beta, `v4.0.0-beta` - for people that want to continue using and contributing it (the last version there is `4.0.0-beta.32`).

And honestly - to downgrade RN, you will need to provide some information in your PR. Reopen it with enough information to give context and reasons about this change. 

There is a guide in the repo that talks a bit about how to contribute: 
https://github.com/aksonov/react-native-router-flux/blob/master/CONTRIBUTING.md

Issue #3176, opened 2018-08-13
Stack of same scene not working anymore after update from 4.0.0-beta-28 to 4.0.1


fdobre:
### Version
Tell us which versions you are using: 

- react-native-router-flux v4.0.1
- react-native v0.55.4

### Expected behaviour

Having some scenes like:

    <Scene key="articles">
        <Scene
          key="home"
          component={ChannelsScreen}
          sceneStyle={styles.screen}
          renderRightButton={<ProfileIcon />}
          onRight={() => {}}
        />

        <Scene
          key="article"
          component={ArticleContainer}
          title={I18n.t('back').toUpperCase()}
          sceneStyle={styles.screen}
          leftButtonIconStyle={styles.goBackImageStyle}
          hideDrawerButton
          tintColor={styleConstants.colorBlack}
          onLeft={() => Actions.pop()}
          leftButtonImage={backButtonLeftArrowImage}
          renderRightButton={() => <View />}
        />
     </Scene>

when calling  `Actions.push('article', newParams });` multiple times, a new article screen should be opened on top of the oldest one increasing the stack items.

### Actual behaviour
old article screen is replaced with a new one

This used to work in `4.0.0-beta-28`

I tried to  put `Scene key` article inside a stack or even transform articles one from scene to Stack but that doesn't seem to work.

Is there any possibility in 4.0.1 to create a stack of same scenes with different info?

### Steps to reproduce
For non-obvious bugs, please fork this component, modify Example project to reproduce your issue and include link here.
1.
2.
3.

Asinox:
The same thing here!

**Version**
Tell us which versions you are using:

react-native-router-flux 4.0.1
react-native 0.56.0

abhimanyuPathania:
Same here

react-native-router-flux 4.0.0-beta-40
react-native 0.51.0

This is probably because of the 'less pushy' breaking change in react-navigation V2
https://reactnavigation.org/blog/#navigateroutename-in-stacknavigator-is-less-pushy


But `react-navigation` still exposes older push actions. Here the change seems to be in `Actions.push`, and I could not find any alternative to it that uses old push behaviour.
https://github.com/aksonov/react-native-router-flux/commit/d61960c6bf0139150a327a161b3d9374c6d6772f#diff-2c2bc1d792d94da8f451749a3b894588R840

I think `Actions.push` should default to older push behaviour and maybe we can have an `Actions.navigate` for the new behaviour.

[...]

PR #3245, opened 2018-09-06
fixes Actions.reset when getParent returns null

ChillkroeteTTS:
Hey,

I currently don't have time to dig deep into this library so I'm not 100% sure if it makes sense what I did but what happened to me after using master branch to get the fix for #3176 was the following:

JSX was:
```
<Router key='router'>
  <Modal onRequestClose={() => { }} key='modal'>
    <Lightbox key='lightbox' hideNavBar>
      <Scene key='root' hideNavBar>
        <Scene key='splash' component={Splash} hideNavBar/>
        [...]
      </Scene>
    </Lightbox>
    <Scene key='configuration' component={InitConfiguration} hideNavBar/>
  </Modal>
</Router>
```

The console output from this.state() inside the _reset_ method called from 'initConfiguration'-screen was:
```
{ key: 'StackRouterRoot',
   routes:
     [ { params: { isInitialNavigation: true },
          routeName: 'configuration',
          key: 'id-1536232996974-7' } ],
      index: 0,
      isTransitioning: false }
```

Due to this combination the getParent([...]) call inside the _reset_ method returned null which lead to a 'null is not an object' error while accessing the 'key' property.

### DIFF
--- a/src/navigationStore.js
+++ b/src/navigationStore.js
@@ -919,11 +919,11 @@ class NavigationStore {
 
   reset = (routeName, data) => {
     const params = filterParam(data);
-    const { key } = getParent(this.state, routeName);
+    const parent = getParent(this.state, routeName);
     this.dispatch(
       StackActions.reset({
         index: 0,
-        key,
+        key: parent ? parent.key : null,
         actions: [
           NavigationActions.navigate({
             routeName,


aksonov:
Looks good!

daviscabral:
Good catch - thanks for fixing that.

PR #3265, opened 2018-09-13
Closes #3163: Bring TabBarTop and TabBarBottom back to LegacyTabs

### LINKED ISSUE #3163: Tabs flag for swiping not working
### Version
Tell us which versions you are using: 

- react-native-router-flux 4.0.0-beta.40
- react-native 0.55.4

### Expected behaviour

When the flag swipeEnabled is set to true, user should be able to swipe through the tabs

I have my tabs layout like this:

```
 <Tabs key="tabbar" tabs={true} swipeEnabled={true} showLabel={false} hideNavBar style={{ marginTop: -50 }} type={ActionConst.RESET}>
   <Scene key="contacts" component={Contacts} hideNavBar type={ActionConst.RESET}/>
   <Scene key="chats" component={Chats} hideNavBar initial={true} type={ActionConst.RESET}/>
   <Scene key="profile" component={Profile} hideNavBar type={ActionConst.RESET}/>
 </Tabs>
```

The same goes for not be able to hide the tab Bar Component which is at the bottom by default.

### Actual behaviour

Swiping is not available. Only works switching via bottom nav

daviscabral:


aksonov:
So, should we release new version now?

daviscabral:
Give me a few minutes - I need to get one more fix pushed - related to one bug I brought to master with the reset/refresh change I did a few days ago.

aksonov:
Okey, I've just got error tabBarComponent is read-only when set `legacy` for `TabBar` (I added `legacy` to Example). I fixed that but got another error within `backToInitial` processing, so I disabled `backToInitial` for `legacy` at this moment, feel free to fix. Now it works but with warning about `jumpToIndex` is deprecated - looks like using of old navigators was bad idea...

daviscabral:
Yeah - sorry about that - I noticed it right after, and I thought I was going to be able to fix in time, but I had to stop to work. 

I have other PR with that and other stuff in development.

PR #3295, opened 2018-09-26
Fix example and add more info to README

daviscabral:
Probably we should warn users about this breaking change in the `createReducer` API.

This PR fixes the broken example after the circular fix and adds more information in README.


aksonov:
Sorry for breaking change, my mistake. I've just published 4.0.4 with old API. preserving of API is very important for us and we should try to keep it stable. Could you remove 'new' API from examples for this PR? Thanks.

daviscabral:
Sure thing.

aksonov:
LGTM

Issue #3293, opened 2018-09-26
hideTabBar not working in nested scenes


xxmuaddib:
### Version 4.0.3
Tell us which versions you are using:

- react-native-router-flux v4.0.3
- react v16.4.2
- react-native v0.55.4

### Expected behaviour

hideTabBar should hide the tab bar in scenes.

### Actual behaviour

hideTabBar doesn't hide the tab bar.

### Steps to reproduce
For non-obvious bugs, please fork this component, modify Example project to reproduce your issue and include link here.
1. Create <Stack>
2. Inside Stack create <Tabs>
3. Inside Tabs create <Stack>
4. Inside Stack create <Scene hideTabBar={true} ... />

### Reproducible Demo

```
<Stack key="root">
  <Stack key="Test" type={ActionConst.RESET} title="Test" hideNavBar>
    <Tabs key="TestTabs" title="TestTabs">
      <Stack key="TestTab" icon={TestIcon}>
        <Scene key="TestScene" component={TestComponent} initial />
        <Scene key="TestScene2" component={TestComponent2} hideTabBar={true} />
      </Stack>
    </Tabs>
  </Stack>
</Stack>
```

I have seen that there is a closed issue on this topic but it didn't help, and 4.0.2 changelog included this as fixed bug. That is why I opened another issue mentioning v4.0.3
And in the API documentation the <Scene> has no hideTabBar prop.

EthanDM:
I am also having this issue. I was previously on 4.0.0-beta.28 which used React Navigation 1.5 and worked fine. I am now trying to upgrade to 4.0.5 to support iPhone XS Max etc.

I'm not sure why there is a closed issue on this - it's definitely a bug.

BS-Tao:
I have the same issue.hideTabBar doesn't work in nested stack scenes.

aksonov:
@daviscabral Could you look?

daviscabral:
Sure.

daviscabral:
> I'm not sure why there is a closed issue on this - it's definitely a bug.

If that's not working in the current version - we are going to look into - but v3 and beta versions are not supported (that is stated in the README, issue template and in several other issues already opened). Sorry if that was not clear. I am looking into that now.

forsen:
We're having this issue on Android as well (ref ios label).

blackbing:
@forsen  Confirm that `hideTabBar` in` Stack` is worked, but it still not worked in `Scene` Props.

[...]

PR #3315, opened 2018-10-11
support hideTabBar on nested scenes

forsen:
In React Navigation 2.x the feature to hide the tab bar was changed, ref: https://github.com/react-navigation/react-navigation-tabs/issues/19

This pull request makes it possible to hide the tab bar from nested scenes, as long as the scene(s) is wrapped in a `<Stack hideTabBar />` component (note that the `hideTabBar` prop have to be set on the stack component, not the scene component). 

I have not thoroughly tested this PR, but it works for our simple case. 

I'm not sure if this is the correct approach / solution to the problem, but we'd like to solve this issue some time soon. Let me know if I can be of any other assistance. 

ref: #3293

### DIFF
--- a/src/navigationStore.js
+++ b/src/navigationStore.js
@@ -324,11 +324,14 @@ function createNavigationOptions(params) {
 
     // currect dynamic navigation params has priority over static scene params
     // but taking them into account only if they are explicitly set (not null or undefined)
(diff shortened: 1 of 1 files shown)


og24715:
Please bump version. 😂

PR #3319, opened 2018-10-15
handle transparency and inherited duration

craftgear:
this PR solves two problems:

1.  backgoundColor black out when 'transparent' specified
2. `duration` is not considered as Router attributes

I encounter these problems when I use RNRF without transition and with transparent background.

related: https://github.com/react-navigation/react-navigation/issues/2713#issuecomment-338414647


craftgear:
Sorry to bother, this isn't necessary.

I find I can use `transitionConfig` attribute.

PR #3320, opened 2018-10-17
Added missing indicatorStyle to TypeScript definition

patw0929:
Tabs component has `indicatorStyle` actually, but it is missing in TypeScript definition (`index.d.ts`).


aksonov:
LGTM!

PR #3361, opened 2018-11-19
prevent navigating twice when clicking a button quickly

iherem:
i using react-native-router-flux@4.0.6 and this bug still the same.

### DIFF
--- a/src/navigationStore.js
+++ b/src/navigationStore.js
@@ -888,11 +888,13 @@ class NavigationStore {
   };
 
   dispatch = (action) => {
-    if (this.externalDispatch) {
-      this.externalAction = action;
-      this.externalDispatch(action);
-    } else if (this._navigator) {
-      this._navigator.dispatch(action);
+    if (action.routeName !== this.currentScene) {
+      if (this.externalDispatch) {
+        this.externalAction = action;
+        this.externalDispatch(action);
+      } else if (this._navigator) {
+        this._navigator.dispatch(action);
+      }
     }
   };
 
--- a/src/navigationStore.js
+++ b/src/navigationStore.js
@@ -888,6 +888,15 @@ class NavigationStore {
   };
 
   dispatch = (action) => {
+    if (action.params && action.params.forceReplace) {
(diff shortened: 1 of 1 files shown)


shx1545:
When I returned, I immediately entered the component again and the component was mounted twice.

aksonov:
@iherem Could you check my comment above?

src/navigationStore.js
```diff
@@ -888,11 +888,22 @@ class NavigationStore {
...
+      }
+      return;
+    }
+    if (action.routeName !== this.currentScene) {
+      if (this.externalDispatch) {
+        this.externalAction = action;
+        this.externalDispatch(action);
+      } else if (this._navigator) {
+        this._navigator.dispatch(action);
+      }
```
pistonsky: What if I have a scene which accepts params, and can navigate to itself, but with a different param? For example, imagine app where you read blog posts. The scene where you read a certain blog post is called "Post". Inside of it, you see a hyperlink which basically takes you to another blog post, so the route name doesn't change. If this change gets approved, all those links stop working.

You need to add a logic that checks how many milliseconds are passed since user last navigated. If it's really too fast, (less than... 500ms? I'm a fast guy, I tap fast, so maybe 300ms?) - then cancel navigating.

But I wouldn't bother. What was the problem you had?

Bang9:
I think this is not a bug.. if this is a bug, react-navigation will fix this first
debounce is user's job, see the example code

```
import { TouchableOpacity } from 'react-native';
import _ from 'lodash';

export default class DebouncedTouchable extends Component {
  static defaultProps = {
    onPress: () => {}
  };

  constructor(props) {
    super(props);
    this.debouncedOnPress = _.debounce(
      () => {
        this.props.onPress();
      },
      300,
      { leading: true, trailing: false }
    );
  }

  render() {
    return (
      <TouchableOpacity {...this.props} onPress={this.debouncedOnPress}>
        {this.props.children}
      </TouchableOpacity>
    );
  }
}

```

PR #3380, opened 2018-12-16
Fix/update example for react-native

s-h-a-d-o-w:
Trying to use createReducer caused:

example_error

Also, instead of getSceneStyle there is now sceneStyle which expects a Style.


aksonov:
LGTM!

PR #3381, opened 2018-12-16
Add basic test for react-native example

s-h-a-d-o-w:
Simply checks whether rendering the example throws, to avoid something like https://github.com/aksonov/react-native-router-flux/pull/3380 in the future.

(I haven't used CircleCI before though, so I don't know whether the config will work like that and for some reason, I can't try it out on the branch of my fork...)


aksonov:
LGTM, let's try to merge that

PR #3389, opened 2018-12-30
add hideNav prop

VB10:


VB10:
add props "hideNavBar" for tabs component (it's without props  generate warning rn ts project)

PR #3401, opened 2019-01-16
Update API.md

Lerosen:
update Documentation for `Actions.create(...)` to include the props and wrap by arguments.
Found this out while trying to set a sceneStyle on my Router via `getSceneStyle`. This prop is ignored if the navigator is set directly on the router, this might be a bug. 
If i supply the cardstyle to `Actions.create( ... , { cardStyle: { ... } }` it is working fine so the Documentation should mention this.


aksonov:
LGTM

PR #3438, opened 2019-03-02
fix: add SceneProps to TabsStatic

douglas-mason:
Per documentation, `<Tabs>` should receive the same props as `<Scene>` and is the same as `<Scene tabs>`, however, the prop type doesn't currently include SceneProps.

This adds SceneProps to TabStatic as is currently done with ModalStatic and DrawerStatic.


aksonov:
LGTM

PR #3442, opened 2019-03-05
Fixed scene title prop type

boris-burgos:
The scene title prop accepts a string or a function as value, but the type is only string.
A function returning string type has been added to the scene title type.


aksonov:
LGTM

PR #3506, opened 2019-07-04
Make NavigationStore an explicit parameter (opt-in)

radex:
Right now `Actions` is the library's API. Being a singleton, it causes some issues for me:

- I want to be able to explicitly pass this as context to components - e.g. for mocking in tests
- I want to be able to have more than one Routers (and therefore multiple NavigationStores) in the app -- for multiple window support on iPadOS 13 and macOS

My solution is to remove the `./navigationStore` singleton internally, and instead pass it explicitly as a parameter. Externally, `Actions` is exported and works as it did, but as an option, there's now a parameter to `Router` to pass user's own instance of `NavigationStore`


aksonov:
LGTM, could you please resolve conflicts?

radex:
@aksonov done

aksonov:
Thanks! Sorry for the delay

PR #3574, opened 2019-10-01
Optimize images

arnaud-zg:


aksonov:
I don't understand why we need that

PR #3612, opened 2020-01-23
Update jest to 25.1.0

friederbluemle:
This also updates the transitive `@jest/reporters` dependency, which updates `istanbul-reports`. The newer version of `istanbul-reports` **drops** the dependency on `handlebars` (an older version of which was causing security warnings).

It also updates the version of Node.js used on CircleCI from 8 to 10.


aksonov:
Thanks, could you update node version for circleci to pass tests?

friederbluemle:
Sure, I updated Node.js to 10 on CircleCI! 👍

PR #3808, opened 2021-12-12
Update to opencollective-postinstall

stianjensen:
The https://github.com/opencollective/opencollective-cli package is
deprecated, and
https://github.com/opencollective/opencollective-postinstall is a more
lightweight alternative, with no transitive dependencies.


stianjensen:
Another option now that requires no packages is to use the 'funding' field in package.json:
https://blog.opencollective.com/beyond-post-install/

I can update to do that instead, if you prefer.

